PR #4, opened 2015-05-20
Sort the dates.

ralphbean:


Cydrobolt:
It works like magic! :+1:
Thanks for the help!

PR #12, opened 2015-05-26
Migrate to python-memcache

Cydrobolt:
Since `python-pylibmc` is not available in EPEL7, we are migrating to `python-memcached`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,14 +3,13 @@ møte - Fedora meetbot log wrangler
 
 ### About
 
-møte allows the Fedora community to search and explore IRC meetings. 
+møte allows the Fedora community to search and explore IRC meetings.
 More information on meetings can be found [here](https://fedoraproject.org/wiki/Meeting_channel?rd=Fedora_meeting_channel)
 
 ### Using møte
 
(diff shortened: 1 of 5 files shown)


ralphbean:
Looks good to me!  :+1:

sontek:
Make sure to update the README and remove the information about getting libmemcached.  

pylibmc does have some nice features that python-memcached doesn't have like consistent hashing so that when you rebalance nodes you don't have to rebalance all the keys at the same time.  Not sure if you are interested in that feature or not.

Cydrobolt:
@sontek Gotcha. I'm changing the README to remove references to `libmemcached`.
I don't think I had planned a method to rebalance nodes with pylibmc, but if that comes into play, I think we might have to deal with the issue as-is :(

ralphbean:
> pylibmc does have some nice features that python-memcached doesn't have like consistent hashing so that when you rebalance nodes you don't have to rebalance all the keys at the same time. Not sure if you are interested in that feature or not

That is really cool.

However, we almost never add new memcached nodes -- that would be a super rare event for us.

PR #24, opened 2015-06-14
fedmsg-based cache updater

### LINKED ISSUE #2: Listen on the fedmsg bus to update memcached when a new meeting ends

ralphbean:
It is a pretty simple script that should be run by systemd with ``systemctl start mote-updater``.

This should fix #2.


mote/fedmsg_consumer.py
```diff
@@ -13,8 +14,21 @@
...
-# Currently a placeholder
-pass
+target = '.meetbot.meeting.complete'
+
+def main():
+    for _, _, topic, msg in fedmsg.tail_messages():
+        if not topic.endswith(target):
+            continue
+
+        print("A meeting just ended!  Let's sleep for 2s to dodge a race.")
```
Cydrobolt: Shouldn't we use `print "x" ` instead of print("x"), since we are on Python 2?

files/mote-updater.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=cache updater for mote
+After=network.target
+Documentation=https://github.com/fedora-infra/mote
+
+[Service]
+ExecStart=/usr/bin/mote-updater
```
Cydrobolt: Where in the spec does the fedmsg-consumer file get copied to /usr/bin/mote-updater?

Cydrobolt:
Thanks for adding the Fedmsg consumer scripts!
I added a couple line comments, but everything else looks good. I haven't had the chance to build it and test it out in a VM, but it looks pretty good so far :)

mote/fedmsg_consumer.py
pypingou: ``print()`` has been backported to py 2.7 and maybe even to 2.6 (but not sure there)

Cydrobolt: Oh, I thought an `import from __future__` was needed to obtain `print()`, although I could be wrong.

Cydrobolt:
This error occurs when building the RPM:

```
Checking for unpackaged file(s): /usr/lib/rpm/check-files /builddir/build/BUILDROOT/mote-0.2.1b1-1.el7.centos.x86_64
error: Installed (but unpackaged) file(s) found:
   /usr/bin/mote-updater
    Installed (but unpackaged) file(s) found:
   /usr/bin/mote-updater
```

Cydrobolt:
I was able to get it to build by adding mote-updater to the file list, although I'm still getting an error when trying to run the mote-updater service.

http://ur1.ca/mtpey

Cydrobolt:
Hmm, it seems like updating all the dependencies to the versions specified in `setup.py` fixes the issue, perhaps we need to change the versions in the setup to conform to the versions packaged by Fedora, or perhaps we can remove those dependency listings from the `setup.py` file.

Cydrobolt:
Hmm, yup. We could remove the specific versions and make it work. Although that allows the script to run, the config is still not found when the updater script is ran. Perhaps we need to specify the env var so that mote can find the config located in `/etc/mote`

ralphbean:
> Perhaps we need to specify the env var so that mote can find the config located in /etc/mote

Good call.  I think a89148fcb4fc will do that.  Can you test again?

Cydrobolt:
037a442e1204df4932a2fe12bab7543ac7dfbceb should fix the version annoyances.

Cydrobolt:
The service now runs correctly, albeit it does not seem to be correctly updating the cache. Perhaps it is due to the print() function being called?

http://ur1.ca/mtsai

Cydrobolt:
We need to figure out what `audit2allow` did to fix the Fedmsg networking, but this PR seems to be good to merge. Once we spin up a new RPM, I can test it again in my VM and figure out what I did.

`httpd_can_network_connect` seems likely, however.

PR #30, opened 2015-06-17
Convert POST endpoints to GET.

ralphbean:
POST doesn't actually make sense here.  It is for when you want to
*change* some data on the server.  Here, we are just requesting that the
server send us some logs -- so that should be a *GET*.

Furthermore, POSTs make things unnecessarily hard to debug, since you can't
easily copy around URLs from the developer console to play with.

I also modified those flask endpoints to return real ``404`` status codes.
If we want to markup 404 pages, we can add a flask 404 handler to catch and
decorate those.


Cydrobolt:
Looks good to me.

PR #34, opened 2015-06-17
Sort search results by most-recent-meetings first.

### LINKED ISSUE #32: search should sort results by most-active-entries first
if you search for just 'design' you see that the first two results it returns have only old entries.. but the third entry is the one you want with the most recent stuff.

ralphbean:
Fixes #32.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Dependencies:
  - Python 2.7.x
 
 `pip` Dependencies:
- - Install using `pip install -r requirements.txt`
+ - Install using ``python setup.py develop`
 
 Running møte:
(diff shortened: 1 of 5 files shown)


Cydrobolt:
This looks pretty good! Have you tested this yourself this?

ralphbean:
> Have you tested this yourself?

I have!  It's working :)

Cydrobolt:
@threebean maybe we should also add the dependency to the spec? :)

ralphbean:
> @threebean maybe we should also add the dependency to the spec? :)

Yes!  I forgot the .spec file was included in the repo.

Cydrobolt:
I added it for you :)

PR #39, opened 2015-06-18
Move insertion of MOTE_FOLDER_PATH to *before* the first import of config.

ralphbean:
The ``from util import ...`` line, does an import of config in the util.py file.


Cydrobolt:
No idea why this bug is only occurring when ran in mod_wsgi, but this fix works great.

PR #50, opened 2015-07-30
Loosen requirements on date parsing regex.

### LINKED ISSUE #25: Bug with FAmSCo meetings
There is a bug with FAmSCo meetings, they seem to return an error. Perhaps someone accidentally edited one of the file names or file formats in a way that broke mote. This error is caused more specifically by a mislabled date. Maybe we can ignore the error, or show a link to the -raw.

ralphbean:
And also complain loudly when we cannot parse a date.

This is should fix #25.

We were failing to parse some dates and returning 'Date Undefined' which was then being cached and causing problems in other parts of the stack which tried to parse that string.

### DIFF
--- a/mote/soke.py
+++ b/mote/soke.py
@@ -44,9 +44,9 @@ def memcached_dict_add(dictn, key, val):
 
 def get_date_fn(filename):
     # Return a meeting's date from a filename.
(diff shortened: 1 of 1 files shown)


Cydrobolt:
That's cool, thanks for fixing this!

PR #51, opened 2015-07-30
Rename "View Log" to just "View".

### LINKED ISSUE #47: Rename "View Log" button to just "View"
"View Log" is confusing when it is present for both the Log and Minutes.

ralphbean:
This should fix #47.


Cydrobolt:
:+1:

PR #61, opened 2015-08-13
mappings: Add g11n category as well as name mappings

pnemade:
Added g11n category as well as name mappings


Cydrobolt:
Looks good

PR #65, opened 2015-09-10
Improved README.md

vivekanand1101:
Added link to [config.py](https://github.com/fedora-infra/mote/blob/master/files/config.py) and [mote](https://github.com/fedora-infra/mote/tree/master/mote) and  removed unnecessary back quote. Also, replaced unnecessary long url for [name_mappings](https://github.com/fedora-infra/mote/blob/master/name_mappings.json) and [category_mappings](https://github.com/fedora-infra/mote/blob/master/category_mappings.json)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Dependencies:
  - Install using `python setup.py develop`
 
 Running møte:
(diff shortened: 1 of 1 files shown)


Cydrobolt:
Looks pretty good!

PR #68, opened 2016-01-14
Add new name mappings, add new category and categorize new names

jflory7:
This pull request adds several new name mappings for different sub-projects or groups in Fedora, such as the Fedora Council, Fedora CommOps, Fedora EMEA Ambassadors, and the Fedora Magazine.

Additionally, it also alphabetizes all existing lists to help improve navigation and user-friendliness for those browsing to find meetings.

Also, minor adjustment to README to change `yum` to `dnf`.

PR #69, opened 2016-01-14
Drop "Fedora Board", use "Fedora Council"

jflory7:
Copy and pasted from commit:
This commit removes official mention of the Fedora Board from møte as it is now defunct in lieu of the new Fedora Council. Since the Council replaces the Board, the name mapping for the Fedora Council includes the keyword for the Fedora Board, so all of the previous meetings in the history of the upper-level deciding body of Fedora are grouped and clustered together.

This aims to make it easier to search and navigate through older and current meetings relating to higher-level Fedora leadership.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,12 @@ møte - Fedora meetbot log wrangler
 ### About
 
 møte allows the Fedora community to search and explore IRC meetings.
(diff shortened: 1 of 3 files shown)


jflory7:
Duplicate pull request, please refer to #68.

PR #72, opened 2016-01-16
Fix regex to catch correct segment of filename

cydrobolt:
This issue was noted by `robyduck` on IRC. 
Specifically, this team was not showing up in search due to its symbols in its filename. 

See https://meetbot.fedoraproject.org/sresults/?group_id=fedora-it_-_incontro_periodico_della_comunita_italiana&type=team

However, it seems like it is now showing up through search. @puiterwijk, did you launch a hotfix to fix this or did it fix itself?

This patch loosens certain aspects of the regex and adds better heuristics for the filename ending, to prevent filenames with `-` symbols from interfering with the date and file extension detection mechanisms.

### DIFF
--- a/mote/soke.py
+++ b/mote/soke.py
@@ -44,7 +44,7 @@ def memcached_dict_add(dictn, key, val):
 
 def get_date_fn(filename):
     # Return a meeting's date from a filename.
-    m = re.search(".*?[\-\.]([0-9]{4}\-[0-9]{2}\-[0-9]{2})[\-\.].*", filename)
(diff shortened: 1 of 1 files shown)


ralphbean:
That's a lot of regex!

@cydrobolt do we have a way to test this?

@puiterwijk can you tell us if you applied this in prod?  If so, that would solve it for us.

cydrobolt:
@ralphbean I've tested this on my own, and it seems to work fine.

You can test the regex on https://regex101.com/ -- it's extremely helpful

ralphbean:
Cool, :+1: here.  :)

cydrobolt:
I'm going to merge this. I'll also write some tests for this specific type of situation so that we don't run into it again. thanks ralph!

PR #74, opened 2016-01-17
Add Fedora LATAM Ambassadors, fix CommOps name

jflory7:
This pull request adds the Fedora Latin American Ambassadors to the møte browse page to easily find all of the meetings hosted by the LATAM community. For the aliases, I searched through møte to find all aliases used by LATAM Ambassadors in the past - there's a lot! I included all of the ones I found as aliases.

Additionally, when browsing møte, I noticed for CommOps that the friendly name wasn't being displayed. The name used in category_mappings was not the same as name_mappings. This was corrected in this PR.


jflory7:
Ran tests using `run_tests.sh` script. Result came out "OK".

cydrobolt:
Hmm. I can't seem to find the `fedora-commops` group. However, I was able to find the `commops` group. Could you send me a link for the location of `fedora-commops`?
image

category_mappings.json
```diff
@@ -1,6 +1,6 @@
 {
-    "Ambassadors": ["famna", "emea_ambassadors", "#fedora-ambassadors"],
-    "Community Operations": ["commops"],
+    "Ambassadors": ["famna", "emea_ambassadors", "fedora-latam", "#fedora-ambassadors"],
+    "Community Operations": ["fedora-commops"],
```
cydrobolt: This should say `commops`

name_mappings.json
```diff
@@ -13,7 +13,7 @@
     },
     "fedora-commops": {
```
cydrobolt: This should also be updated to `commops`, and `commops` should be removed from the aliases here`

jflory7:
@cydrobolt Commit 05c1e7b should fix the issue you pointed out above. :)

cydrobolt:
Great, thanks @jflory7

ralphbean:
@cydrobolt, did you want to merge this one?

cydrobolt:
@ralphbean ah, yes. Sorry for the delay!

@jflory7 can you check to make sure all the aliases/groups you are adding exist in `mote`? Some of the aliases you have (e.g `CommOps`) do not exist in the system. See: https://meetbot.fedoraproject.org/sresults/?group_id=CommOps&type=team

All team/channel names are case sensitive, so we need to make sure that they exactly match. You might also want to rebase some of the commits together, to help organise these changes better. 

Thanks for the contributions!

jflory7:
@cydrobolt I had tried to anticipate potential aliases that might be used in the future for CommOps, but none of them actually existed yet. For the time being, I have truncated the aliases down to the only other alternative alias that exists for CommOps as of right now.

I tried playing around with rebasing, but I couldn't get it to work in the way I was wanting. I'll take some more time to familiarize myself with this part of git, but I also didn't want to delay your time with me figuring it out, so I committed ffb1de7 in the meanwhile.

cydrobolt:
Ah, okay. I see. No worries. I can rebase it for you later.

PR #76, opened 2016-01-21
Add aliasing

cydrobolt:
- Add aliasing
 - Change year to 2016
 - Remove redundancy in obtaining friendly name

### DIFF
--- a/files/test_config.py
+++ b/files/test_config.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright © 2015 Chaoyi Zha <cydrobolt@fedoraproject.org>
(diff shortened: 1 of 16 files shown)


files/test_config.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright © 2015 Chaoyi Zha <cydrobolt@fedoraproject.org>
+# Copyright © 2016 Chaoyi Zha <cydrobolt@fedoraproject.org>
```
ralphbean: I think technically, this is supposed to be ``2015-2016``

cydrobolt: Ah, you're right! I should change this.

ralphbean:
Looks good to me.  :)

Issue #79, opened 2016-01-31
Use alises as a way to merge groups


jflory7:
Many different sub-projects and groups in Fedora are inconsistent with the meeting names they use when starting meetings with zodbot. Sometimes this is unintentional, sometimes it is just not being aware that consistent meeting names are important.

Until recently, my understanding was that aliases were a solution to this by "merging" different group names together under a singular, common group name. However, it does not seem to work this way.

## Solution
Proposed solution is to have groups defined in the `category_mappings.json` file or similar. Groups could then have multiple aliases for what their meeting names could be, and then all meetings with those meeting names are lumped together via aliases.

#### Use Case
* Group that is meeting: Fedora LATAM Ambassadors
* Group name in møte: `latam_ambassadors`
* Potentially used meeting names:
 * `fedora_latam_ambassadors`
 * `latam_ambassadors`
 * `latam`
 * `fedora_latam`
 * `fedora_latam_ambassadors_meeting`

And so on. If any of these meeting names were used, via aliases, they would all be sorted together under the meetings for "Fedora LATAM Ambassadors".

PR #82, opened 2016-02-08
Add FZUG (#fedora-zh)

alick:
Please let me know if there is the right way to add a new channel.

### DIFF
--- a/category_mappings.json
+++ b/category_mappings.json
@@ -13,5 +13,6 @@
     "Project Leadership": ["fedora_council"],
     "Quality Assurance": ["fedora-qa"],
     "Release Engineering": ["releng"],
-    "Websites": ["fedora-websites"]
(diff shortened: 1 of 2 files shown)


cydrobolt:
Thanks @alick :)

cydrobolt:
It seems like most of your meetings occur in `#fedora-zh`, not categorized under a specific team/meeting name. We're still working on clustering these meetings into one. See #79

alick:
@cydrobolt yeah, all the meetings happen in `#fedora-zh`. That's for the local community in China.

I thought the alias part is functional and I added various used meeting names there. Guess we'll stick to one meeting name in the future.

cydrobolt:
The alias does help propagate certain changes (such as the friendly name), but you would still have to search for "fedora-zh" to find your meetings until they are bundled together (ticket 79)

Thanks :)

PR #85, opened 2016-02-17
Add new fedora-mktg alias

jflory7:
Adds a new alias for the `fedora-mktg` team. Ran the tests script and all checked out.


cydrobolt:
Thanks @jflory7

PR #87, opened 2016-02-20
Add latest meetings section, streamline some code

### LINKED ISSUE #86: Add "latest meetings" to the frontpage
Often, I go to mote *shortly* after I participated in a meeting.  It would be nice if the frontpage had a (small) set of links to the last 3 or 4 meetings, so I could quickly find it without having to search.

cydrobolt:
Maintains a list of latest meetings in the past 24 hrs by triggering a datagrepper fetch each time `soke` is run.

Latest meetings are cached along with other cache data.

Fixes #86


mote/soke.py
```diff
@@ -105,9 +107,14 @@ def run():
                 except:
                     pass
 
+    # fetch latest meetings using datagrepper for the past 24 hours
+    latest_meetings = get_latest_meetings()
```
ralphbean: This needs exception handling around it, in case datagrepper is down.  If datagrepper dies, perhaps just let ``latest_meetings = []`` (graceful failure).

ralphbean:
Looks good!  Just needs that one fix around the datagrepper call as far as I can see.

mote/latest_meetings.py
```diff
@@ -0,0 +1,36 @@
...
+seconds_delta = 86400;
+topic = 'org.fedoraproject.prod.meetbot.meeting.complete';
+
+url_template = "https://apps.fedoraproject.org/datagrepper/raw?delta={}&topic={}".format(seconds_delta, topic)
+
+def get_latest_meetings():
+    # fetch meetings from the last day using datagrepper
+    last_day_raw = requests.get(url_template)
+
+    if last_day_raw.status_code != 200:
```
ralphbean: This will work.  But FYI, you can do ``if not bool(last_day_raw):`` to catch more edge cases.

There are more "good" [status codes](http://http.cat) than just "200".

ralphbean:
:+1:, looks good!

PR #88, opened 2016-02-22
Hash on a link now scrolls to tm properly

rdp1070:
Now when the page loads with a hash, it animates down to the appropriate location.

### DIFF
--- a/mote/static/js/single-log.js
+++ b/mote/static/js/single-log.js
@@ -32,6 +32,11 @@ function loadLogContents() {
         var markup;
         markup = res;
         $('.logdisplay').html(markup);
+        var line = document.getElementsByName(location.hash.split('#')[1]);
+        $line = $(line);
+          $('html, body').animate({
+            'scrollTop' : $line.offset().top
+        });
         return true;
     });
     return true;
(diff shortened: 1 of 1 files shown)


cydrobolt:
@rdp1070 thanks for submitting a PR!

I think you're on the right track with this issue, but it doesn't look like the patch works completely. If you run the mote server and test with an anchor inside the `pre`, it seems to scroll to the first couple of lines no matter which line is specified in the hash.

rdp1070:
I am sorry, I am kind of new to this. I have python installed but I am not sure how to run the mote server. I used the test file and it passed, so I guess it's not broken, but the functionality may be off like you said.

cydrobolt:
@rdp1070 no worries! You need to install the dependencies from the `requirements.txt` (`pip install -r requirements.txt`) then copy the configuration file over. `cp files/config.py mote/config.py`

Once the dependencies and configuration files are properly set, you can run mote by running `python mote/runmote.py`.

@jflory7 told me you were on Windows, so you'll have to make sure both `pip` and `python` are in your path before you can run those commands.

Take a look at this: https://www.python.org/downloads/release/python-279/

rdp1070:
Thanks! I got the environment setup. Now it's time to go at this bug. 
PR before my night is done.

cydrobolt:
@rdp1070 I fixed some issues with line offsetting and merged your PR.
Thanks for the contributions! :+1:

PR #90, opened 2016-03-03
Add datagrepper URL as option

### LINKED ISSUE #89: Add datagrepper URL as a configuration variable
The datagrepper URL as referenced in `latest_meetings.py` (https://github.com/fedora-infra/mote/blob/master/mote/latest_meetings.py#L22) should be a configuration variable to allow users to specify their datagrepper instance and allow `meetbot.stg.fp.o` to work correctly. When the datagrepper prod URL is used in the stg mote instance, an SSL error is shown:

```
SSLError: hostname 'apps.fedoraproject.org' doesn't match either of '*.stg.fedoraproject.org', 'stg.fedoraproject.org'
```

The stg mote instance should use apps.stg.fp.o/datagrepper rather than apps.fp.o/datagrepper

cydrobolt:
Fix #89

### DIFF
--- a/files/config.py
+++ b/files/config.py
@@ -25,6 +25,8 @@ meetbot_fetch_prefix = "http://meetbot-raw.fedoraproject.org"
 # Time (in seconds) after which the log/meeting cache expires
 cache_expire_time = 60 * 60 * 1
 
+# datagrepper URL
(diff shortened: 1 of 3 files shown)


ralphbean:
LGTM!  :dango:

cydrobolt:
:dango:

Issue #97, opened 2016-05-10
No logs


puiterwijk:
There are no logs in /var/log/httpd/error_log on 500's and no indication of where the logs are in the SOP.

Please add logging so server errors can be debugged.

cydrobolt:
The httpd log file should contain logs; however, they should also be available if you run `service httpd status`

I'll check on this soon.

puiterwijk:
service log:
May 11 01:15:08 value01.phx2.fedoraproject.org systemd[1]: Starting The Apache HTTP Server...
May 11 01:15:09 value01.phx2.fedoraproject.org httpd[28036]: [Wed May 11 01:15:09.053645 2016] [env:warn] [pid 28036] AH01506: PassEnv variable HOSTNAME was undefined
May 11 01:15:09 value01.phx2.fedoraproject.org systemd[1]: Started The Apache HTTP Server.

[root@value01 httpd][PROD]# cat error_log
[root@value01 httpd][PROD]# 


homepage:
500

Sorry, it seems like something went wrong.

If this was unexpected, please open a ticket on our issue tracker.



This has NOT been resolved. Please fix ASAP.

cydrobolt:
Looks like the function server_error_500(e): in the mote init.py should be logging e to stderr.

cydrobolt:
See #99 @puiterwijk

puiterwijk:
Looks like this still isn't fixed.
Mote again threw 500 responses, with an empty error_log.

cydrobolt:
We never pushed this update afaik. I'll try to build it and ping you.

puiterwijk:
@cydrobolt Note that we are currently frozen, and value01 freezes...
Also, is there any specific reason this was never built and updated, as now we again crashed without having any idea as to why.

cydrobolt:
I'm not sure why this was not built and updated. It's possible that it was lost in the mix. I can built it and package it, then we can deploy the fix once the freeze is over.

cydrobolt:
I've built mote 0.6.1 on Koji: https://koji.fedoraproject.org/koji/taskinfo?taskID=18652391

cydrobolt:
It's been deployed. Closing again.

PR #98, opened 2016-05-10
Add logging for mote mod_wsgi

cydrobolt:
#97 
@puiterwijk, can you review this and see if this could work?

### DIFF
--- a/files/mote.conf
+++ b/files/mote.conf
@@ -10,6 +10,9 @@ WSGIPythonOptimize 1
 
 WSGIScriptAlias / /usr/share/mote/mote.wsgi
 
+ErrorLog ${APACHE_LOG_DIR}/mote.log
+LogLevel warn
+
 <Location />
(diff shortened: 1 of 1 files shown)


puiterwijk:
Is there any reason why it would be a separate logfile?
The standard apache error log would be most ideal since it'd automatically get rotation rules and sync.

cydrobolt:
@puiterwijk hm, you're right. What's the default `LogLevel`? Maybe that's all we need to fix the issue. Or perhaps no errors have been raised in the app so far?

puiterwijk:
Well, the app was throwing 500's for a long time this afternoon and I couldn't find any logs.

cydrobolt:
Ah. Weird. The default `httpd` LogFile has been reliable for me in the past. I'm not sure what could've changed

puiterwijk:
The homepage. https://meetbot.fedoraproject.org/

PR #99, opened 2016-05-17
Output traceback from 500 handled errors onto stderr

cydrobolt:
Uses the traceback module to output Flask's caught errors onto `stderr`.

### DIFF
--- a/mote/__init__.py
+++ b/mote/__init__.py
@@ -29,6 +29,7 @@ except:
 
 import flask, random, string, json, util, re
 import dateutil.parser, requests, collections
+import logging
 
 from six.moves import html_parser as html_parser_six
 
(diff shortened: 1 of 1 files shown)


puiterwijk:
-1.
Writing to stdout/stderr is suboptimal for web services.
I would suggest you to instead use the python logging module.

cydrobolt:
@puiterwijk is this better deb5b17?

puiterwijk:
This is a reasonable start of the code, yes. But as far as I know this only initializes the logging code and does not actually log anything yet.

cydrobolt:
@puiterwijk try running the app. Flask seems to automatically detect the logger and logs to it when the errors are caught.

puiterwijk:
Ah, okay, I didn't understand you were depending on that behavior.
:+1: then.

mote/__init__.py
```diff
@@ -68,6 +69,9 @@
 name_mappings = map_name_aliases(json.loads(name_mappings))
 category_mappings = json.loads(category_mappings)
 
+logging_format = '%(asctime)-15s %(message)s'
+logging.basicConfig(format=logging_format)
+logger = logging.getLogger('tcpserver')
```
puiterwijk: You might want to make this logging.getLogger(__name__)

puiterwijk:
:shipit: 可

PR #103, opened 2016-10-03
Add Fedora Badges team mapping

jflory7:
This pull requests adds a mapping for the Fedora Badges team, which will begin meeting regularly on Wednesdays at 11:00 UTC.


jflory7:
@cydrobolt That was fast – thanks for the quick review. :smile: 

:clapper:

cydrobolt:
😃

PR #108, opened 2016-11-11
Add different icons in search results to differentiate types of groups

### LINKED ISSUE #106: Group types are confusing
It's confusing when you're looking for a team in m0te and you get the team's IRC channel as one group and the actual team as another group. It's easy to gloss over / not notice the 'type' text field and assume you're getting two results for the same thing:
 
screenshot from 2016-11-02 15-32-48

One suggestion to make this less confusing would be to mark the group's type with an icon, which at a glance could make it clear - especially if it's right next to the group name - what type of group it is more easily than a separate text field:

bitmap

It would also make sense to continue this scheme on the group details pages, eg to carry the icon over to pages like this:
https://meetbot.fedoraproject.org/sresults/?group_id=fedora-design&type=team

The icons used in the quick mockup above are from Font Awesome - 
- team - 'users'
- channel - 'comments'

cydrobolt:
Fixes #106

### DIFF
--- a/mote/static/css/mote.css
+++ b/mote/static/css/mote.css
@@ -250,6 +250,14 @@ a {
     word-wrap: break-word;
 }
 
+.select2-results__option h4 {
+    display: inline;
+}
+
+.select2-results__option .group-icon {
+    margin-right: 0.5em;
+}
+
 .latest-meetings-header {
     margin-top: 2em;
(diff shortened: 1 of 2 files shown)


pypingou:
Cool, :+1: for me :)

PR #114, opened 2017-02-17
Reverse date sorting on group pages

### LINKED ISSUE #109: Years listed in descending order but months in ascending order
Hi!
One little niggle with the UI for viewing a team's list of meeting minutes - the years are stacked most recent to least recent, with 2016 on the top. However, the months are least recent to most recent order - so if it's dec 2016, you have to scroll way down to view the notes for the month even tho it's the current month. I'd suggest flipping the months so the most recent month is always at the top; then the list will consistently display most recent notes at the top and least at the bottom. Right now it's out of order bc of the month order.

cydrobolt:
Fixes #109

### DIFF
--- a/mote/__init__.py
+++ b/mote/__init__.py
@@ -317,7 +317,7 @@ def sresults():
     sorted_dates = list(groupx_meetings.keys())
 
     try:
-        sorted_dates.sort(key=dateutil.parser.parse)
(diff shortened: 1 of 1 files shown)


pypingou:
👍 for me :)

cydrobolt:
:+1:

Issue #115, opened 2017-04-16
Latest meeting button is broken


cydrobolt:
The latest meeting function expects to find the latest meeting at the top of the bottom of the array of meetings. However, 9d27661a533302f5454f42e227f5a1bb27411ff1 reversed the order of the months, and the latest meeting is now present at the top of the array. 

When the latest meeting button is pressed, the incorrect date is pulled up. Instead of pulling the date from the bottom, it should be pulled from the top of the array.

PR #116, opened 2017-04-21
Fix latest meeting button

cydrobolt:
#115

### DIFF
--- a/mote/__init__.py
+++ b/mote/__init__.py
@@ -345,7 +345,7 @@ def sresults():
         type = group_type,
         avail_dates = avail_dates,
         meetbot_location = config.meetbot_prefix,
-        latest_meeting = (avail_dates.items()[0][1].items()[0][1])[0]
+        latest_meeting = avail_dates.items()[0][1].items()[0][1][0]
     )
 
 
--- a/mote/__init__.py
+++ b/mote/__init__.py
@@ -29,7 +29,7 @@ except:
 
 import flask, random, string, json, util, re
 import dateutil.parser, requests, collections
-import logging
+import logging, soke
 
 from six.moves import html_parser as html_parser_six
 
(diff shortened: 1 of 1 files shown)


cydrobolt:
@pypingou could you review this?

mote/__init__.py
```diff
@@ -29,7 +29,7 @@
 
 import flask, random, string, json, util, re
 import dateutil.parser, requests, collections
-import logging
+import logging, soke
```
pypingou: This doesn't seem to be used in the diff here

mote/__init__.py
```diff
@@ -345,7 +345,7 @@ def sresults():
         type = group_type,
         avail_dates = avail_dates,
         meetbot_location = config.meetbot_prefix,
-        latest_meeting = next(reversed(avail_dates.items()[0][1].items()[-1][1]))
+        latest_meeting = (avail_dates.items()[0][1].items()[0][1])[0]
```
pypingou: I don't think you need the brackets, do you?

cydrobolt: You're right, I don't think we need the parentheses.

mote/__init__.py
cydrobolt: The import is used elsewhere in the file, but it's never caused problems even though it was never imported before. 

pyflakes was complaining about the missing import, so I added it. I'm not sure why it was working before even though the import was missing. Maybe you have a better idea of why?

pypingou: Because it seems to be only used in Except clauses, so if you never faced the exception, it was never called :)

cydrobolt: Interesting, thanks for clearing that up for me :)
We should probably add the import. I remember someone mentioning that random 500s happened occasionally. Maybe those could have been related to this issue.

pypingou: We should definitively keep the import since otherwise we are creating an error inside a exception handling :)

cydrobolt:
I removed the unnecessary parentheses. Thanks @pypingou!

PR #122, opened 2019-06-12
Adapt code to be python3 compatible and easier to test

keszybz:
The tests now pass, also when run under pytest. But I didn't do any "real" testing. Somebody
who knows how this is supposed to be set up should take a look.


files/test_config.py
```diff
@@ -22,9 +22,10 @@
 
 Crawler Configuration
 '''
-cwd = os.getcwd()
+root = os.environ['TEST_DATA']
```
pypingou, line 25: I don't see the use of the environment variable before and the code here assumes that it is always present

run_tests.sh
```diff
@@ -14,55 +14,32 @@
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 #
 
+[ -n "$PYTHON" ] || PYTHON=python2
```
pypingou, line 17: defaults to python2?

keszybz: This was written a while back... Should python2 support just be dropped now?

files/test_config.py
keszybz: It is set in `run_tests.sh`. See the commit message for explanation.

run_tests.sh
keszybz: I looked over this patch set, and it still seems fine. The changes are minimal in the sense that python2 is still used by default. If you want, I can add another commit on top that either switches to python3 by default or completely drops python2 support.

nirik:
If we want to move this app to python3, thats just fine with me. The only other thing on value01 is zodbot, but we can move it to python3 also and rhel8 at the same time.

jwflory:
Thanks @keszybz!

@pypingou / @nirik, is there anything more needed to merge this PR? I can rebase #124 after.

nirik:
Seems fine to me. I'll leave the last word for @pypingou :)

jwflory (changes requested):
For Travis CI tests to work, you need to change the Python version from 2.7 to 3.8:

https://github.com/fedora-infra/mote/blob/37ab930e355b9846cdaf7764318e44b524cf87b9/.travis.yml#L2-L3

keszybz:
> For Travis CI tests to work, you need to change the Python version from 2.7 to 3.8:
> 
> https://github.com/fedora-infra/mote/blob/37ab930e355b9846cdaf7764318e44b524cf87b9/.travis.yml#L2-L3

Done.

jwflory:
@keszybz Thanks! So, now Travis actually attempts to run the test suite with Python 3.

However, it looks like some files are not in the right path. Is there an obvious fix for this? (I have not dug into the code changes in this PR)

https://travis-ci.org/github/fedora-infra/mote/builds/732466317#L286

If it is an easyfix, it would be nice to clean that up. If it is more involving, I propose we open an issue and encourage others to help us improve our tests with us.

keszybz:
The tests pass now.

jwflory:
Excellent! This should be ready to merge. :smile: 

@pypingou Are you able to help us wrap this one up?

pypingou:
I'd recommend the git history should be clean a bit, some of the commits can likely be squashed

keszybz:
> I'd recommend the git history should be clean a bit, some of the commits can likely be squashed

I think the history *is* clean. It consists of many small steps, logically separate. But if you tell me what exactly needs to be squashed, I will squash it.

pypingou:
Well, all the py2 to py3 commits could be squashed. Keep removing the spec file in its own commit. Make the changes to the tests in their own commit as well.

keszybz:
> Well, all the py2 to py3 commits could be squashed. Keep removing the spec file in its own commit. Make the changes to the tests in their own commit as well.

Nah, that's pretty bad. Some of those commits are done automatically by applying 2to3 with a script, and others are done manually.

pypingou:
>  Nah, that's pretty bad. Some of those commits are done automatically by applying 2to3 with a script, and others are done manually.

I'd say how they are done is a technical detail that does not really matter.

Anyway, it looks like we disagree and I'm not going to fight for or against this so I'm stepping out of this review.

keszybz:
I don't think the commit structure matters. Whoever owns the repo and merge the PR can squash the commits as they see fit. I don't mind at all, it's just that the commit structure was done purposefully in a way that allows the same sequence of operations to be applied on other python projects. Here the code is very simple, but in general separating the steps out makes the whole thing easier.

jwflory:
I am happy to merge this but I do not have access to this repo.

jwflory (approved):

jwflory:
Huzzah, thanks @pypingou for helping facilitate this, and thanks @keszybz for working on this (especially after 1+ year of making this PR!!). I'm going to merge this so we can work on some other changes and maybe repay some other tech debt on Mote over the next year.

Merging! :clapper:

jwflory:
Also, I am going to take a hybrid approach for the merge strategy discussed above. Going to opt for a squash+merge, but since the commits were not squashed in this Pull Request, we can always point a future Python developer to this PR to see how Møte did its Py2 to Py3 transition.

PR #123, opened 2020-02-24
Fix syntax highliting in meeting logs

mymindstorm:
| Before | After |
| ------ | ----- |
| image | image |


jwflory:
Wow, this is a super easy bugfix and would readability so much easier. I'd love to see this make it to production!

cydrobolt:
Looks good to me. I haven't been active in a while and don't have commit access anymore, so someone else will have to take this change to production.

pypingou:
> Looks good to me. I haven't been active in a while and don't have commit access anymore, so someone else will have to take this change to production.

That's something we can fix :)

pypingou:
@cydrobolt welcome back :)

Zlopez (approved):

PR #130, opened 2020-10-03
Created CODE_OF_CONDUCT.md

antriksh123:


t0xic0der (commented):
You are requested to kindly stick to the tried-and-tested Contributor Covenant Code of Conduct than to come up with something new. You can find one here. https://github.com/fedora-infra/mote/community/code-of-conduct/new?template=contributor-covenant.

Also, please wait before the issue is validated. (We don't know if the issue is worth solving or not)

jwflory:
Hi! Actually, we cannot add this because we need to use the Fedora Code of Conduct. It is also under revision, to be shared later in 2020.

Let's revisit a Code of Conduct for the Fedora Infra GitHub repos later this year. It is important work, but it will quickly become outdated. I think it is better to send people looking for a Code of Conduct instead of giving them the wrong one. :slightly_smiling_face:

jwflory:
For the time being, I am going to close this Pull Request and we can revisit with discussion in the issue.

PR #133, opened 2020-10-03
Updated footer in base.html

antriksh123:
updated the footer so that it shows in mobile view


jwflory:
Hi @antriksh123! I am not familiar with the CSS used here.

Could you please share a before/after screenshot of what this change looks like? It will be easier to understand what this change does then.

antriksh123:
I know that @nasirhm and that is why i removed the class. But which theme does it use from bootswatch because i myself use bootswatch.

jwflory (approved):
OK, simple enough to me. Thanks for adding context @nasirhm. I am going to go ahead and merge this, thanks @antriksh123! :clapper:

nasirhm:
> I know that @nasirhm and that is why i removed the class. But which theme does it use from bootswatch because i myself use bootswatch.

I'm not sure on that one but we're working on updating the UI.

PR #135, opened 2020-10-10
Containerization

t0xic0der:
Packaged the dependencies with Python 3.8 reference

Versions => Python 3.8 have complete support


requirements_venv.txt
```diff
@@ -0,0 +1,29 @@
...
+python-memcached==1.59
+python-openid==2.2.5
+python-openid-cla==1.2
+python-openid-teams==1.1
+python3-openid==3.2.0
+requests==2.24.0
+six==1.15.0
+soupsieve==2.0.1
+urllib3==1.25.10
+Werkzeug==1.0.1
```
jwflory, line 29: Out of curiosity, was there a particular reason to make an additional `requirements_venv.txt` in addition to `requirements.txt` in the root directory? It might get confusing if they get out-of-date.

venv/bin/Activate.ps1
```diff
@@ -0,0 +1,241 @@
+<#
+.Synopsis
```
jwflory, line 2: It is convention to add the `venv/` directory to the `.gitignore`. I don't think we need to commit all of these files. :sweat_smile:

jwflory:
Hey @t0xic0der, thanks for opening this PR! :+1: I am all for better virtualenv management. But I think using a Python virtualenv management tool like [Pipenv](https://pipenv.pypa.io/en/latest/) or [Poetry](https://python-poetry.org/) would be better here. They have a better user experience, and take some of the headaches out of using a virtualenv on different platforms and operating systems.

Would you be up for exploring either one of these two virtualenv management tools? I have used Pipenv for a long time and I am comfortable with it, but I know kids these days seem to be hyped on Poetry. :sunglasses: :grin:

requirements_venv.txt
t0xic0der: The `requirements.txt` available from the get-go describes the bare minimum versioning of the dependencies needed for this project, while the `requirements_venv.txt` talks especially about the specific version of dependencies which made their way inside the virtual environment and are frozen (for now).

venv/bin/Activate.ps1
t0xic0der: The entry of `venv/` was there on `.gitignore`. :laughing: I kinda added them even when the `requirements_venv,txt` could have just done fine. I'll remove them.

t0xic0der:
I have been using virtualenv for so long that I can hardly remember when it was the first time I started off using it - so being a package vet, virtualenv did bring a lot of nuisance but I was able to get over them. But you are right - this process is more about fighting with the system and less on making the system work for you. :sweat_smile: 

Yes, Pipenv is something that even I am comfortable with and Poetry shows hope. (Only hope, because it kinda makes things way too easy for folks that some necessary nuances are abstracted)

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: python
 python:
   - "3.8"
 # command to install dependencies
-install: "pip install -r requirements.txt"
+install: "pipenv install"
```
jwflory, line 5: You might try this to make Travis CI play more nicely with Pipenv:

```suggestion
install: "pip install $(pipenv lock --requirements)"
```

Pipfile
```diff
@@ -0,0 +1,31 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+arrow = ">=0.7.0"
+beautifulsoup4 = ">=4.3.2"
+flask = ">=0.10.1"
```
jwflory, line 11: Unrelated to this Pull Request… but wow we should get Møte up to Flask 1.1.x.

nasirhm: @jwflory If you could create a ticket for it, I'll be more than happy to port it :)

t0xic0der:
Once the checks are cleared, I'd get into resolving the conflicts. (Why is it taking this long?)

Pipfile
t0xic0der: There you go man https://github.com/fedora-infra/mote/issues/138, :smile:

t0xic0der: But fulfilling the requirements would get you 1.1.x by default as the requirements are leniently laid down. It simply asks for a version above 0.10.1. We should more likely update the code to incorporate versions greater than. say, 1.0 maybe.

jwflory: +1. Thanks for opening an issue to track this! :+1:

nasirhm:
@t0xic0der The following PR will fix issue: https://github.com/t0xic0der/mote/pull/1

Once it's merged, Tests will work fine.

t0xic0der:
Thanks for the PR @nasirhm. :smile:

nasirhm:
Welcome, Please resolve the Conflicts with `.travis.yml` and fire up the Build :)

t0xic0der:
Done. @nasirhm. Waiting on the check.

.travis.yml
```diff
@@ -1,14 +1,10 @@
...
-    - language: python
-      python: 3.8
-      cache: pip
-      install: "pip install -r requirements.txt"
-      script: sh run_tests.sh
+language: python
+python:
+  - "3.8"
+# command to install dependencies
+install: "pip install $(pipenv lock --requirements)"
```
nasirhm, line 5: How about using: `pipenv install` to ensure the Build works fine and then we can figure out a way to use pip instead ?

t0xic0der: Yes. @nasirhm. That seemed to have worked.

.travis.yml
```diff
@@ -1,14 +1,10 @@
...
-dist: bionic
-
-jobs:
-  include:
-    - language: python
-      python: 3.8
-      cache: pip
-      install: "pip install -r requirements.txt"
-      script: sh run_tests.sh
+language: python
```
nasirhm, line 1: As @jwflory worked on modernizing the `.travis.yml` file for the CI, Please add: 

```
os: linux
dist: bionic

jobs:
  include:
     - language: python
       ...
```

nasirhm (changes requested):
Please modify `.travis.yml` and start a build on CI and ensure the tests work fine.

.travis.yml
t0xic0der: There. Done. :)

nasirhm (commented):
LGTM

nasirhm (commented):
LGTM, What do you think @jwflory ?

.travis.yml
```diff
@@ -1,14 +1,18 @@
...
+     - language: python
+
+python:
+  - "3.8"
+
+# command to install dependencies
+install: "pipenv install"
+
+# command to run tests
+script: sh run_tests.sh
```
jwflory, line 15: The Travis config appears to have reverted to an older status. This Pull Request needs to be rebased on the latest changes in `master`. Once the Travis config is fixed up, we can merge this and the docs PR next.

jwflory (changes requested):
One fix remaining before this can be merged. Once the Travis file is cleaned up, we can merge and follow up with the other open Pull Requests.

nasirhm (changes requested):
Please add the updated `requirements.txt` where, 
`python-openid>=2.2.5` is removed
In scope of this PR: https://github.com/t0xic0der/mote/pull/1

jwflory (approved):
The changes look good to me. I think this is ready to merge. :grinning: 

Merging. :ocean:

Issue #134, opened 2020-10-10
[Documentation] Rework README.md to make include Python 3 specific instructions


t0xic0der:

nasirhm:
Working on it.

Issue #140, opened 2020-10-20
[Bug] : config.json while memcached being False


nasirhm:
While trying to run the Application with copying the `config.py` file to `mote/` , It runs the application along with attaching a debugger to the WSGI application.

While accessing it via the browser, It was caused by the following issues (Stacktrace of the Debugger):
```
Traceback (most recent call last):
  File "/Infra/mote/mote/util.py", line 61, in get_json_cache
    with open(config().json_cache_location, mode='r') as json_store:
FileNotFoundError: [Errno 2] No such file or directory: '/var/cache/httpd/mote/cache.json'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Infra/mote/mote/__init__.py", line 98, in get_cache_data
    res = util.get_json_cache(meeting_type)
  File "/Infra/mote/mote/util.py", line 76, in get_json_cache
    raise RuntimeError("Cache not found.")
RuntimeError: Cache not found.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Infra/mote/mote/util.py", line 61, in get_json_cache
    with open(config().json_cache_location, mode='r') as json_store:
FileNotFoundError: [Errno 2] No such file or directory: '/var/cache/httpd/mote/cache.json'
```
Why are we storing the `cache.json` file in a root-level directory `/var/cache/httpd/cache.json` if we're using the `python runmote.py` to run the development build ?

jwflory:
_Discussed in [2020-10-25 meeting](https://youtu.be/gpLPDgYRIvA)_.

---

This looks like some cruft still hanging in from when `memcached` was a more useful dependency. @nasirhm agreed to tackle this one for the current sprint, ending on Sunday, 8 November 2020. Will be completed after #138 is finalized.

PR #141, opened 2020-10-20
Documentation to run møte on Py3

nasirhm:
Modified Documentation for Python 3
Usage of Pipenv to facilitate with the creation of Virtualenv and installation of packages inside it.

Fixes: #134 

Dependent on PR: #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,17 +15,22 @@ møte organises and serves meetings as a drop-in replacement, without needing an
 ### Using møte
 
 Dependencies:
- - Optional: `memcached` (`sudo dnf install memcached`)
- - Python 2.7.x
+ - `memcached` (`sudo dnf install memcached`)
+ - Python 3.8.X
+ - Pipenv
 
 Installation:
- - Optional: create virtual environment
- - `pip install -r requirements.txt`
- - Install using `python setup.py develop`
+ - `pipenv install` will create a new virtual environment and install the requirements. `
 
 Running møte:
+ - Start `memcached` by typing `memcached` in the terminal and pressing Return (Enter) key.
  - Edit [config.py](./files/config.py) with necessary information and place the `config.py` in [`mote/`](./mote)
+ - Modify the copied `config.py` to use memcached by setting `use_memcached = TRUE`
+   - You can do it by running the following command in command line too: `sed -i '/use_memcached = / s/False/True/' mote/config.py`
+ - Enter the Pipenv's newly created virtualenv's shell with `pipenv shell`
  - Run `python runmote.py`
+
+**Note :** Make sure `memcached` is running in the background.
 
 Running tests:
  - `sh run_tests.sh`
--- a/README.md
+++ b/README.md
@@ -15,12 +15,13 @@ møte organises and serves meetings as a drop-in replacement, without needing an
 ### Using møte
 
 Dependencies:
- - `memcached` (`sudo dnf install memcached`)
- - Python 3.8.X
- - Pipenv
+ - [`memcached`](https://memcached.org) (`sudo dnf install memcached`)
+ - Python 3.8.x
+ - [Pipenv](https://docs.pipenv.org/) (`sudo dnf install pipenv`)
 
-Installation:
- - `pipenv install` will create a new virtual environment and install the requirements. `
+Virtual Environment setup:
+ - `pipenv install` will create a new virtual environment and install the requirements.
+ - `pipenv shell` will activate Python's virtual environement's shell.
 
 Running møte:
  - Start `memcached` by typing `memcached` in the terminal and pressing Return (Enter) key.


t0xic0der:
You are right. #135 needs to be merged to fully realize the changes made here.

I followed the instructions provided here for the `pull/135` branch and found them to be working just fine.

Good work, @nasirhm.

t0xic0der (approved):
LGTM. Followed the instructions provided here and it works just how it is supposed to.

nasirhm:
> LGTM. Followed the instructions provided here and it works just how it is supposed to.

Thank you for the review, one thing to take care of in the Development experience in the future would be to include a sample `config.json` related file in order to make `memcached` an Optional dependency and ensuring it doesn't cause any errors in while keeping the usage of `memcached = False`
Resulted in me creating issue #140 for further research.

README.md
```diff
@@ -15,18 +15,23 @@ møte organises and serves meetings as a drop-in replacement, without needing an
...
- - Python 2.7.x
+ - `memcached` (`sudo dnf install memcached`)
+ - Python 3.8.X
+ - Pipenv
 
 Installation:
- - Optional: create virtual environment
- - `pip install -r requirements.txt`
- - Install using `python setup.py develop`
+ - `pipenv install` will create a new virtual environment and install the requirements. `
```
jwflory, line 23: For Pipenv, I usually recommend two steps:

```sh
# Install dependencies
$ pipenv install

# Activate Python virtualenv
$ pipenv shell
```

README.md
```diff
@@ -15,18 +15,23 @@ møte organises and serves meetings as a drop-in replacement, without needing an
 ### Using møte
 
 Dependencies:
- - Optional: `memcached` (`sudo dnf install memcached`)
- - Python 2.7.x
+ - `memcached` (`sudo dnf install memcached`)
+ - Python 3.8.X
+ - Pipenv
```
jwflory, line 20: A link to the Pipenv docs would be a good idea to add here. Also, this is packaged in Fedora too. `sudo dnf install pipenv`

README.md
```diff
@@ -15,18 +15,23 @@ møte organises and serves meetings as a drop-in replacement, without needing an
...
 
 Running møte:
+ - Start `memcached` by typing `memcached` in the terminal and pressing Return (Enter) key.
  - Edit [config.py](./files/config.py) with necessary information and place the `config.py` in [`mote/`](./mote)
+ - Modify the copied `config.py` to use memcached by setting `use_memcached = TRUE`
+   - You can do it by running the following command in command line too: `sed -i '/use_memcached = / s/False/True/' mote/config.py`
+ - Enter the Pipenv's newly created virtualenv's shell with `pipenv shell`
  - Run `python runmote.py`
 
+**Note :** Make sure `memcached` is running in the background.
```
jwflory, line 33: It is out of scope for this Pull Request, but maybe we can use `docker-compose` or `podman-compose` to make this easier to set up and run for new developers.

nasirhm: I was thinking to something similar, I don't think if we would need `docker-compose` / `podman-compose` , we can proceed with a simple `Dockerfile` too. What do you think ?

nasirhm: Here's an issue to keep track of it: https://github.com/fedora-infra/mote/issues/142

README.md
nasirhm: Fixed: https://github.com/fedora-infra/mote/pull/141/commits/8bfdd02674b734fa256248bf2429a00fddf28a11

README.md
nasirhm: Fixed: https://github.com/fedora-infra/mote/pull/141/commits/8bfdd02674b734fa256248bf2429a00fddf28a11

README.md
```diff
@@ -15,12 +15,13 @@ møte organises and serves meetings as a drop-in replacement, without needing an
...
- - Pipenv
+ - [`memcached`](https://memcached.org) (`sudo dnf install memcached`)
+ - Python 3.8.x
+ - [Pipenv](https://docs.pipenv.org/) (`sudo dnf install pipenv`)
 
-Installation:
- - `pipenv install` will create a new virtual environment and install the requirements. `
+Virtual Environment setup:
+ - `pipenv install` will create a new virtual environment and install the requirements.
+ - `pipenv shell` will activate Python's virtual environement's shell.
```
t0xic0der, line 24: An ordered list maybe? (Because they need to be executed one after the other.)

t0xic0der (changes requested):
Consider the changes I suggested before implementing them. (See if they are really needed)

README.md
nasirhm: @t0xic0der Thanks for the review, personally I feel, It's pretty intuitive with unordered lists to follow it step by step and the list representation seems simple as points, Do you think adding an ordered list would make it more simple ?

t0xic0der: Either could work. There are just two points anyway. :-)

jwflory (approved):
The docs here look good to me. Since #135 was merged, we are ready to pull this one in too.

Merging! :clapper: Thanks @nasirhm.

Issue #126, opened 2020-10-03
Add CONTRIBUTING.md


t0xic0der:
In the spirit of getting more contributors to the project, the above document would be appreciated.

antriksh123:
You can close this or review my pull request on CONTRIBUTING.md

jwflory:
I think we need to work with a member of the Community Platform Engineering team on writing this document. There are things I do not know about CPE workflows and how to request reviews. So I think we should partner with someone who can answer the trickier questions about how to contribute here.

t0xic0der:
> I think we need to work with a member of the Community Platform Engineering team on writing this document. There are things I do not know about CPE workflows and how to request reviews. So I think we should partner with someone who can answer the trickier questions about how to contribute here.

This stays open for now and I'll check if there is a mailing list for CPE. If so, I'll reach out to them to see if they can give some pointers regarding this.

jwflory:
> This stays open for now and I'll check if there is a mailing list for CPE. If so, I'll reach out to them to see if they can give some pointers regarding this.

The Infra list is the best place to go for that. :grin: https://lists.fedorahosted.org/archives/list/infrastructure@lists.fedoraproject.org/

jwflory:
_Discussed in [2020-10-25 meeting](https://youtu.be/gpLPDgYRIvA)_.

---

I am taking a first attempt at this. I plan to split off the developer environment section in the README into the CONTRIBUTING doc. From there, it should be easier to on-board new developers who want to work on a Flask project.

Together with @ramyaparimi, we will tag-team this in the second half of the sprint to make sure it is accurate and makes sense from different viewpoints. This should be complete by the end of the sprint on Sunday, 8 November, 2020.

Ank221199:
Hey, I would like to take up this issue

t0xic0der:
Done.

PR #144, opened 2020-10-27
Pipfile: Upgrade env to Python 3.9

jwflory:
This commit raises Møte's supported Python version to Python 3.9. In the
upcoming Fedora 33 release, Fedora 3.9 becomes the default version of
the Python interpreter. So, this follows Fedora's precedent by using the
latest version of upstream Python.

Since we are still early in development, we can be more flexible about
this for now.

Part of #126, but does not complete it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ dist: bionic
 jobs:
   include:
     - language: python
-      python: 3.8
+      python: 3.9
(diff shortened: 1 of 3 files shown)


ramyaparimi:
+1 
I agree with this 👍

ramyaparimi (approved):

Issue #148, opened 2020-11-01
Determine final fate of memcached in Møte


jwflory:
_Discussed in [2020-11-01 sprint meeting](https://youtu.be/7l81smp61R8)_.

---

# Summary

Determine whether memcached dependency is worth keeping in.

# Background

`memcached` is a dependency from the original project work, but after recent difficulties with running it locally, we are wondering how necessary it is. Fedora Infrastructure also retired their memcached service some time ago, so there is no longer an easy way of using this in production either.

In the sprint meeting where this issue came up, we noted that supporting memcached expands the scope for the Møte Modernization Project. We are most effective if we can focus on development. If we also need to run our own infrastructure and support a memcached deployment, that will take away valuable, limited time of the volunteer team to work on and improve Møte.

We need to make a final decision in the coming sprint for whether we strip it out of the project, or whether we explore ways to better integrate and/or find a new way to run this dependency both for development and production.

# Details

There are two options to move forward. @t0xic0der and @nasirhm are tag-teaming this one and will come back with a recommendation for the team by 22 November.

### Keep memcached

1. Better document and support running memcached for local development (preferably with containers)
1. See if we can request an [Amazon ElastiCache service](https://aws.amazon.com/elasticache/memcached/) from CPE Team to run memcached in production
1. Explore other edge cases once we get it running consistently

### Drop memcached

1. Remove unnecessary user-facing config options for memcached.
1. Clean up any Møte code that has strong reliance on memcached, if any

# Outcome

* Less friction for local development with difficult-to-test dependencies
* Think ahead for the future, once we get to a point where we are ready to deploy

nasirhm:
_Discussed in [2020-11-08 sprint meeting](https://youtu.be/dJ995Zl2EH4)_.

---

It was suggested to keep `memcached` in the codebase for legacy and future purposes.
 
For development, to prefer the usage of file based cache would be better idea.

PR #149, opened 2020-11-03
config.py now uses cache file in . directory.

nasirhm:
With our current Fedora Infrastructure, we're using a file-based cache storage
we can do the same for our development experience.
while memcached being turned off,
It was trying to write to a file in / root directory resulting in an error (which works fine in the deployed Fedora-infra).

This commit modifies the location of the cache file to create it in the current directory.

It would help developers, who are unable to setup memcached
and would like to use file based cache for development.

Related Issues: #148 #140 
Fixes: #140


t0xic0der:
@nasirhm : I can use some help regarding the expected change here. :sweat_smile:

nasirhm:
> @nasirhm : I can use some help regarding the expected change here.

This change works fine for now and you can use it by turning off the memcached setting :).

It'll be storing the cache in `./cache.json` file in the current directory.

files/config.py
```diff
@@ -48,7 +48,7 @@
 use_memcached = False # Use a memcached store for greater performance
 
 # JSON cache store location
-json_cache_location = "/var/cache/httpd/mote/cache.json"
+json_cache_location = "./cache.json"
```
t0xic0der, line 51: Just as good as promised. :smile: 

image

t0xic0der (approved):
Nice work, @nasirhm.

nasirhm:
Thank Youu @t0xic0der for the review, :D

nasirhm:
_Discussed in [2020-11-08 sprint meeting](https://youtu.be/dJ995Zl2EH4)_.

---

It is suggested, that it is a good idea to get it merged as it provides a better development experience for new developers.

cc: @ramyaparimi and @HiraTariq-01 please provide a review on this PR.

HiraTariq-01 (approved):
Good to go! :D

ramyaparimi:
+1 from me.
working as expected!

ramyaparimi (approved):

nasirhm:
I think, we're good to merge this one, as currently in Fedora Infrastructure, we're using a [separate and modified `config.py`](https://pagure.io/fedora-infra/ansible/blob/master/f/roles/mote/templates/config.py)

Once the CI runs successfully, we're good to merge it. :D

t0xic0der:
I'll make a merge when the CI check completes. :smile:

nasirhm:
Thank you for merging it @t0xic0der

PR #150, opened 2020-11-05
Docker configuration

### LINKED ISSUE #142: Creation of Container Image
While migrating mote app to Python 3. 

For setting up the development environment, You have to install `memcached` and `pipenv` in your host machine.

It would be a good idea to containerize the app to run inside an isolated container with `podman` or `docker` in order to make sure, the developer doesn't have to install packages in their host machine.

**Path 1** : Do we want to go with a plain `Dockerfile` from which we can derive a container image and the developer can build it according to their changes and provide a `Makefile` with rules: `build` and `up` to ease the building and running process of containers with `podman` (if available) else `Docker`

The developer would have to do the following:
```
make new <- to build the container image everytime they modify stuff
make up <- run the built container image.
```

**Path 2**: Another path would be to use `docker-compose` or `podman-compose` with a `docker-compose.yml` file to create and run the container.

The developer would have to do the following:
```
docker-compose up <- to build and run the container
```

A feature for the compose path would be, It'll only recreate containers that have changed.

t0xic0der:
Resolves #142.

Uses `requirements.txt` over `Pipenv.lock` file (to start with).


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3.8-slim
+WORKDIR /mote
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
```
nasirhm, line 4: How about adding:

```
RUN sed -i '/use_memcached = / s/False/True/' mote/config.py
RUN sed -i sed -i 's|json_cache_location = "/var/cache/httpd/mote/cache.json"|json_cache_location = "./cache.json"|' config.py
```

to make sure the user / developer don't have to make any modifications to the project while using the containerized version with the file cache.

nasirhm (commented):
Please share your feedback on the comment.

Dockerfile
t0xic0der: Yep, sounds good. Would make the dev experience a bit more convenient.

t0xic0der:
The changes made in https://github.com/fedora-infra/mote/pull/150/commits/46dc6fb7b4f125117cc26ec9abe9b29dcaa7f40c further makes it convenient to use the JSON file store (which was already claimed to be faster than the `memcached`), hence making the choice redundant. (Assists discussions at https://github.com/fedora-infra/mote/issues/148)

README.md
```diff
@@ -26,11 +26,20 @@ Virtual Environment setup:
...
  - Edit [config.py](./files/config.py) with necessary information and place the `config.py` in [`mote/`](./mote)
- - Modify the copied `config.py` to use memcached by setting `use_memcached = TRUE`
+ - Modify the copied `config.py` to use memcached by setting `use_memcached = TRUE` (Optional)
    - You can do it by running the following command in command line too: `sed -i '/use_memcached = / s/False/True/' mote/config.py`
  - Enter the Pipenv's newly created virtualenv's shell with `pipenv shell`
  - Run `python runmote.py`
 
+OR
+
+Using `docker-compose` to run møte:
```
nasirhm, line 36: As we've modified the Dockerfile with some sed scripting, I think, `docker-compose up` would work fine.

nasirhm (changes requested):
Please fix the changes in the md-docs.

README.md
t0xic0der: On it.

README.md
```diff
@@ -26,11 +26,20 @@ Virtual Environment setup:
...
+ - Start `memcached` by typing `memcached` in the terminal and pressing Return (Enter) key.
```
nasirhm, line 37: We don't need `memcached` in our case, as we're using file based cache.

README.md
```diff
@@ -26,11 +26,20 @@ Virtual Environment setup:
...
+ - Edit [config.py](./files/config.py) with necessary information and place the `config.py` in [`mote/`](./mote)
```
nasirhm, line 38: How about copying it within the container:
```
RUN cp ./files/config.py config.py
```
or adding a If statement, so, if the file exists we wouldn't execute it.
and then the RUN statements for sed.

README.md
```diff
@@ -26,11 +26,20 @@ Virtual Environment setup:
...
+ - Modify the copied `config.py` to use memcached by setting `use_memcached = TRUE` (Optional)
```
nasirhm, line 39: Removal of this line, as we're moving forward with file based storage.

README.md
```diff
@@ -26,11 +26,20 @@ Virtual Environment setup:
...
+   - You can do it by running the following command in command line too: `sed -i '/use_memcached = / s/False/True/' mote/config.py`
+ - Run `docker-compose up`
+
 **Note :** Make sure `memcached` is running in the background.
```
nasirhm, line 43: Removal of this line.

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+RUN sed -i '/use_memcached = / s/False/True/' mote/config.py
```
nasirhm, line 5: Please remove this line, as it asks the config to set `use_memcached` as True from False (By default). And as we're using file based cache, we won't be needing this line.

nasirhm (changes requested):
Good work, please share your thoughts on the changes suggested above.

README.md
t0xic0der: That's been an oversight. I'd look into it.

t0xic0der: (And had we been using `memcached`, we would be better off using the official Docker image https://hub.docker.com/_/memcached)

README.md
t0xic0der: > How about copying it within the container:

You mean making the copy happen with the `Dockerfile`, right?

Dockerfile
t0xic0der: Yes, I was about to say that for we made those changes via https://github.com/fedora-infra/mote/pull/150/commits/46dc6fb7b4f125117cc26ec9abe9b29dcaa7f40c.

README.md
t0xic0der: I think we would want to keep only the
```
RUN cp ./files/config.py config.py
```
in the `Dockerfile` and leave other configurations to the user, for anyway once they are made - they should be copied to be actually some use.

Also, I believe that the destination for this copy should be `./mote/config.py` instead of just `config.py`.

README.md
t0xic0der: Done.

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.8-slim
+WORKDIR /mote
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+RUN cp ./files/config.py ./mote/config.py
```
nasirhm, line 5: What if there's already a modified config.py in the `mote` directory modified by the user, It'll copy the contents of the base again, I think, for now, it would be good to remove it.

docker-compose.yml
```diff
@@ -0,0 +1,6 @@
+version: "3.8"
+services:
+  mote:
+    build: .
+    ports:
+      - "5000:5000"
```
nasirhm, line 6: I would suggest mounting the volume here in `docker-compose.yml` because, it'll keep track of the changes and make sure to rebuild the container in case of any modifications inside the volume.

When we build the container image by COPYing the volume inside Dockerfile, It actually builds into the Docker image with some sort of compression as a Volume specific to that image, resulting in manually rebuilding it when modifying the directory. Adding it to Docker-compose would make sure for us to not do it and makes the development experience much smoother.

(We can do the same with mounting it on the CLI when running it without Docker-compose)

```
volumes:
      - .:/mote
```

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+RUN sed -i sed -i 's|json_cache_location = "/var/cache/httpd/mote/cache.json"|json_cache_location = "./cache.json"|' ./mote/config.py
+EXPOSE 5000
+COPY . .
```
nasirhm, line 8: No need to copy once we have it mapped with docker-compose.

nasirhm (changes requested):

Dockerfile
t0xic0der: Would not be much harm to do so right? As the changes were already made to the file, it would just copy that stuff over but if we remove the 5th line - we make them copy that file again by themselves.

docker-compose.yml
t0xic0der: > ```
> volumes:
>       - .:/mote
> ```

Understood. This is to mount the `/mote` folder as root of the container image.

> When we build the container image by COPYing the volume inside Dockerfile, It actually builds into the Docker image with some sort of compression as a Volume specific to that image, resulting in manually rebuilding it when modifying the directory. Adding it to Docker-compose would make sure for us to not do it and makes the development experience much smoother.

All points noted. But I have two questions regarding these suggested changes.
1. How do we plan on synchronizing the stuff with respect to the changes made to the local repo?
2. Do we rebuild the volume every time to constitute the changes are which are made?

Building an image with mounted volumes takes longer than simply mounting folders as volumes but it is sure to make things consistent and clean.

Dockerfile
t0xic0der: Note to myself: `REMOVE L#5 OF Dockerfile` :smile:

docker-compose.yml
nasirhm: > Understood. This is to mount the /mote folder as root of the container image.

It'll mount the current folder (in which the Docker-compose file lives) on `/mote` of the container Image.

> How do we plan on synchronizing the stuff with respect to the changes made to the local repo?

In a development environment, whenever there'll be changes to the code base, docker-compose will update it inside the container too. (Currently, Not the case as the codebase gets built in the Container image.)

> Do we rebuild the volume every time to constitute the changes are which are made?

We wouldn't have to, Docker-compose will take care of the volume management.

> Building an image with mounted volumes takes longer than simply mounting folders as volumes but it is sure to make things consistent and clean.

as we're using docker-compose, I think, It'll be a good idea to use features that it provides because, it'll ensure, whenever local changes are made, the user wouldn't have to interact with the Docker / Docker-compose CLI interface to rebuild the container on file changes. When the folder as volume is mounted, It'll be updated with in the container too.

P.S: We're not creating a separate volume but asking `docker-compose` to mount the current directory (in which `docker-compose.yml` lives) as a volume with in the container. 

What do you think ?

t0xic0der: That makes sense. I must have misinterpreted stuff previously and so were the questions, as I thought that for every small changes to be reflected in here - we would have to rebuild the image, which is clearly not the case.

t0xic0der:
Take a look. :smile:

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.8-slim
+WORKDIR /mote
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+RUN sed -i sed -i 's|json_cache_location = "/var/cache/httpd/mote/cache.json"|json_cache_location = "./cache.json"|' ./mote/config.py
```
nasirhm, line 5: There's an additional `sed -i` here, The command will be:

```
RUN sed -i 's|json_cache_location = "/var/cache/httpd/mote/cache.json"|json_cache_location = "./cache.json"|' config.py
```

nasirhm (changes requested):
Please fix the requested changed

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.8-slim
+WORKDIR /mote
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+RUN sed -i 's|json_cache_location = "/var/cache/httpd/mote/cache.json"|json_cache_location = "./cache.json"|' ./mote/config.py
```
nasirhm, line 5: Once, This [PR](https://github.com/fedora-infra/mote/pull/149`) is merged, we wouldn't have to modify it.

README.md
```diff
@@ -26,12 +26,16 @@ Virtual Environment setup:
 Running møte:
  - Start `memcached` by typing `memcached` in the terminal and pressing Return (Enter) key.
  - Edit [config.py](./files/config.py) with necessary information and place the `config.py` in [`mote/`](./mote)
- - Modify the copied `config.py` to use memcached by setting `use_memcached = TRUE`
+ - Modify the copied `config.py` to use memcached by setting `use_memcached = TRUE` (Optional)
```
nasirhm, line 29: We wouldn't need `memcached` from now on, as It's by default `FALSE` and It'll default to file based cache.

README.md
```diff
@@ -26,12 +26,16 @@ Virtual Environment setup:
...
    - You can do it by running the following command in command line too: `sed -i '/use_memcached = / s/False/True/' mote/config.py`
  - Enter the Pipenv's newly created virtualenv's shell with `pipenv shell`
  - Run `python runmote.py`
 
-**Note :** Make sure `memcached` is running in the background.
+OR
+
+Using `docker-compose` to run møte:
+ - Edit [config.py](./files/config.py) with necessary information.
```
nasirhm, line 37: How about asking to copy, [config.py](./files/config.py) in [mote/](./mote) first and then using it, as it'll be the case by default ?

As we're doing it [here](https://github.com/fedora-infra/mote/pull/156#discussion_r537334571)

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.8-slim
```
nasirhm, line 1: Add an Environment variable as : `PYTHONUNBUFFERED=1` by adding:

```
ENV PYTHONUNBUFFERED=1
```

In the `Dockerfile`

So, python sends print and log statements directly to `stdout`. If we did not set this, we would not see logs from our container because they would be sent to a buffer.

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+WORKDIR /mote
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+RUN sed -i 's|json_cache_location = "/var/cache/httpd/mote/cache.json"|json_cache_location = "./cache.json"|' ./mote/config.py
+EXPOSE 5000
+COPY . .
```
nasirhm, line 7: How about placing the `COPY` statement before exposing the port, as It'll make the Dockerfile look better. 

What do you think ?

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: "3.8"
+services:
+  mote:
+    build: .
```
nasirhm, line 4: After `build`, you can add the following line:

```
command: python3 runmote.py
```

in `docker-compose.yml`

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+CMD ["python3", "runmote.py"]
```
nasirhm, line 8: I would suggest, removing it from here, as It'll hardcode the `runmote.py` in the container image, and for some cases, a developer might have to run another file.

We can solve it by adding it to `docker-compose.yml`

nasirhm (changes requested):
Please share your thoughts on the review.

Dockerfile
t0xic0der: Does it have a green flag yet?

If so, I can remove the changes made in the L#5 of Dockerfile in this PR to make way for #149.

README.md
t0xic0der: Ah, a README.md change. I'd make this one real quick.

README.md
t0xic0der: Please refrain for exploring project technicalities in the CONTRIBUTING.md.

If at all we are keeping such information, we would want to keep it in README.md.

Dockerfile
t0xic0der: Before L#1 or after L#1?

Dockerfile
t0xic0der: +1 to that. I'd do it.

docker-compose.yml
t0xic0der: Note to myself: `ADD command: python3 runmote.py at L#5`

Dockerfile
t0xic0der: Makes sense.

Also, L#8 does not look very elegant to me.

Dockerfile
nasirhm: After L#1 as we're setting the environment variable in the Image. :)

Dockerfile
t0xic0der: I could use some assistance regarding the reference file locations :sweat_smile:

Dockerfile
```diff
@@ -1,8 +1,9 @@
...
+ENV PYTHONUNBUFFERED=1
 WORKDIR /mote
 COPY requirements.txt requirements.txt
 RUN pip install -r requirements.txt
 RUN sed -i 's|json_cache_location = "/var/cache/httpd/mote/cache.json"|json_cache_location = "./cache.json"|' ./mote/config.py
-EXPOSE 5000
 COPY . .
-CMD ["python3", "runmote.py"]
+EXPOSE 5000
+CMD ["python3", "runmote.py"]
```
nasirhm, line 9: We wouldn't need it once we have the like in Docker-compose: https://github.com/fedora-infra/mote/pull/150/commits/7de8aa6f21528aff693bb49545dae274fc4f74e9#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R5

nasirhm (changes requested):

Dockerfile
t0xic0der: A note to myself `REMOVE L#9 of Dockerfile`

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.8-slim
+ENV PYTHONUNBUFFERED=1
+WORKDIR /mote
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+RUN sed -i 's|json_cache_location = "/var/cache/httpd/mote/cache.json"|json_cache_location = "./cache.json"|' ./mote/config.py
```
nasirhm, line 6: Please remove this line, as the [PR#149](#149) is merged.

nasirhm (approved):
Once it's fixed, we're good to go.

Dockerfile
t0xic0der: L#6 right?

nasirhm (approved):
Looks Good To Me, the CI seems to be failing due to the testcases failing and It doesn't seem to have any effect in the docker configuration.

++ @t0xic0der  for working on it. We're good to merge.

t0xic0der:
@jwflory if you have the push access, please make the merge happen. :)

nasirhm:
@t0xic0der I'm not sure if @jwflory will be able to merge it as the Checks are failing at the moment, we might have to fix them too.

Issue #63, opened 2015-09-10
Fix select2 box to fit in with rest of theme


cydrobolt:
It currently has its own style, since it is generated and isn't affected by `form-control` or other styles imposed on text input boxes.

ramyaparimi:
Opened (PR #153 ) [https://github.com/fedora-infra/mote/pull/153] to fix the issue.

PR #153, opened 2020-11-11
changed the select2 color to match the theme(fixes issue #63)

ramyaparimi:
This PR is created to fix the color of select2 box to match mote blue theme.

This is created for [issue#63](https://github.com/fedora-infra/mote/issues/63).

**Current select2 color**
image

**Proposed select2 color to match the theme:**
image


nasirhm (approved):
LGTM Thank you for working on it @ramyaparimi, 

Here's the diff:
Screenshot from 2020-11-12 12-45-39

t0xic0der (approved):
+1x LGTM from my end too. I didn't see the code differences though but with a change as small as this, I think a screenshot review should suffice.

t0xic0der (changes requested):
Umm... Wait.

Is it possible to introduce changes via inline CSS @ramyaparimi by overriding the classes specified in the `select2.min.css`? I would not want to make changes in the importable CSS file and overriding the classes for making contained changes is a lot more acceptable way of doing things here.

ramyaparimi:
> Umm... Wait.
> 
> Is it possible to introduce changes via inline CSS @ramyaparimi by overriding the classes specified in the `select2.min.css`? I would not want to make changes in the importable CSS file and overriding the classes for making contained changes is a lot more acceptable way of doing things here.

+1
I get what you mean! I made inline changes now. Let me know what you think.

ramyaparimi:
Thank you @nasirhm @t0xic0der for reviewing the PR.
I made added the color in mote.css that would reflect inline. Let me know what you think about it 😃

mote/static/css/mote.css
```diff
@@ -274,3 +274,9 @@ a {
 .latest-meetings-list li {
     list-style-type: none;
 }
+
+.select2-results__option--highlighted[aria-selected] {
+    background-color: #294172 !important;
+    color: white;
+  }
+
```
t0xic0der, line 282: Yes, that's exactly what I was conveying. Making changes to this custom-made CSS file would be much better than adding those changes to an imported one for who knows Select2 might be used somewhere else and not need those theming options.

If the consequence of these changes look the same as before, I think it is ready for a merge.

t0xic0der (commented):
Let me quickly hitch it up locally and approve it once I am done checking.

t0xic0der (approved):
LGTM.

mote/static/css/mote.css
ramyaparimi: This is giving out the same output!! Thanks for reviewing!

nasirhm:
Thank You for working on it @ramyaparimi, It looks awesome, Can you please squash these changes into one to ease the merge ?

ramyaparimi:
umm... so I used the squash option here and accidentally squashed and merged 🤦

PR #161, opened 2021-05-14
Added specfile for Mote 0.7.0 release

t0xic0der:
I have packaged it temporarily in https://copr.fedorainfracloud.org/coprs/t0xic0der/mote/. 

Here's the build status for [Build No. 2184771](https://copr.fedorainfracloud.org/coprs/t0xic0der/mote/build/2184771/).

image

It fails on EPEL8 due to the absence of `fedmsg` package - I believe a blind replacement of `fedmsg` with `fedora-messaging` should work if both are cross-compatible.

Signed-off-by: Akashdeep Dhar <akashdeep.dhar@gmail.com>

### DIFF
--- a/mote.spec
+++ b/mote.spec
@@ -1,0 +1,238 @@
+# There is no need to make our lives so difficult!
+# %{!?python_sitelib: %global python_sitelib %(%{__python3} -c "from
+# %distutils.sysconfig import get_python_lib; print (get_python_lib())")}
+
+Name:       mote
+Version:    0.7.0
+Release:    4%{?dist}
+Summary:    A MeetBot log wrangler, providing a user-friendly interface for Fedora's logs
+
+License:    GPLv2+
+URL:        https://github.com/t0xic0der/mote
+Source0:    https://github.com/t0xic0der/mote/archive/%{version}.tar.gz
+BuildArch:  noarch
+
+BuildRequires: python3-devel
+BuildRequires: python3-pip
+BuildRequires: python3-mod_wsgi
+BuildRequires: python3-flask
+BuildRequires: python3-fedora
+BuildRequires: python3-openid
+# We do not make use of memcached anymore
+# BuildRequires: python3-memcached
+BuildRequires: python3-openid-cla
+BuildRequires: python3-openid-teams
+BuildRequires: python3-requests
+BuildRequires: python3-dateutil
+BuildRequires: python3-beautifulsoup4
+BuildRequires: python3-fedora-flask
+BuildRequires: python3-six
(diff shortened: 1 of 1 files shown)


nirik:
I don't think you can just change fedmsg to fedora-messaging, there's a bit of porting likely I think. 
Perhaps @pypingou or @abompard could comment on that?

And many many thanks for working on this packaging. :)

t0xic0der:
> I don't think you can just change fedmsg to fedora-messaging, there's a bit of porting likely I think.
> Perhaps @pypingou or @abompard could comment on that?
> 
> And many many thanks for working on this packaging. :)

Yes, I got in touch with @siddharthvipul and @mscherer regarding the schema of fedora-messaging and got to know that `fedmsg` and `fedora-messaging` are not cross-compatible and it seems that there would be some codelevel change required before replacing the dependencies. The schema is something to work on, so I thought that the best way forward would be to, first fix the CI here, then ship this version of the package (with Python 3 updates and `fedmsg`) as-is for now and come back to porting support with `fedora-messaging` later.

According to [this](https://src.fedoraproject.org/rpms/fedmsg), `fedmsg` is no longer packaged on EPEL8 and due to the fact that it is one of the dependencies for Mote, my build [fails](https://copr.fedorainfracloud.org/coprs/t0xic0der/mote/build/2184771/) on EPEL8 chroot. Would it be correct to reach out to the current maintainer of `fedmsg` to include support for EPEL8 or do we make do with a [rendition](https://copr.fedorainfracloud.org/coprs/clime/python3-fedmsg-epel8/) of the `fedmsg` package maintained at COPR?

pypingou:
We purposedly do not want to maintain fedmsg in epel8, porting to fedora-messaging isn't really hard and the schema, while necessary, can be done in a second step, which would unblock you. 

https://fedora-messaging.readthedocs.io/en/stable/tutorial/conversion.html has some good doc on converting fedmsg applications to fedora-messaging :)

t0xic0der:
@pypingou Thanks for linking in the documentation.

The `fedmsg_consumer.py` file is predominantly a consumer and not a publisher, which checks for meetings that have recently finished every after 2 seconds and then executes gathering of the meeting information from the datagrepper with `soke.py`. Unfortunately, the documentation does not (yet) cover the consumer side of porting from `fedmsg` to `fedora-messaging`. Could you please link me to some examples for them?

pypingou:
If you are not a publisher it will make things even easier as there will be no need for message schemas.

As for consuming messages the docs is: https://fedora-messaging.readthedocs.io/en/stable/consuming.html

PR #163, opened 2021-05-17
Testing stages 001

copperii:


copperii:
Just for testing and showcasing.

copperii:
py2_test_soke and py2_test_util are currently disabled and have to be ported to python 3 format.

PR #184, opened 2021-10-01
Fixes Issue #174: Add CI for enforcing styles with Black and sorting imports 

### LINKED ISSUE #174: Add CI for enforcing styles with Black and sorting imports
Add CI for enforcing styles with Black and sorting imports

OBITORASU:
## Description
Added a custom GitHub workflow to enforce coding style with black and to sort imports using isort. This workflow triggers on all push and pull requests. It sets up a python 3.8 environment and runs both black and isort on the codebase starting from the root directory. The verbose output of both isort and black can be checked in the GitHub actions log and changes can be introduced with respect to that.

Fixes #174

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,17 @@
+name: run black and isort
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install isort and black
+        run: pip install isort black
+      - name: Run isort
+        run: isort --check --diff .
+      - name: Run black
+        run: black --check --diff .


OBITORASU:
There seems to be conflict in styles between black and isort. 
Here is the log 
```
isort --diff --check .
ERROR: D:\Programming\mote\main.py Imports are incorrectly sorted and/or formatted.
--- D:\Programming\mote\main.py:before  2021-10-01 10:37:24.666703
+++ D:\Programming\mote\main.py:after   2021-10-01 10:38:18.338588
@@ -24,12 +24,8 @@
 import click
 from flask import Flask, abort, jsonify, render_template, request

-from modules.call import (
-    fetch_channel_dict,
-    fetch_datetxt_dict,
-    fetch_meeting_content,
-    fetch_meeting_dict,
-)
+from modules.call import (fetch_channel_dict, fetch_datetxt_dict,
+                          fetch_meeting_content, fetch_meeting_dict)
 from modules.find import find_meetings_by_substring
 from modules.late import fetch_recent_meetings

Skipped 1 files
black --diff --check .
All done! ✨ 🍰 ✨
4 files would be left unchanged.
```

OBITORASU:
The quick fix seems to be setting isort's profile to black for compatibility. Working on the fix right now!

.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
...
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install isort and black
+        run: pip install isort black
+      - name: Run black
+        run: black --check --diff .
+      - name: Run isort
+        run: isort --profile black --check --diff .
```
t0xic0der, line 17: Any reason why we are not sticking with the default profile? I'm curious.

OBITORASU: Yes, the same issue seems to persists if we stick with the default profile I can produce logs for that as well. The only way I got both of them to work in harmony is by setting isort's profile to black.

t0xic0der: Makes sense.

t0xic0der (approved):
LGTM.

@dkirwan, see if there are any changes that you would want to suggest here.

OBITORASU:
Yes I am checking this out.

Issue #180, opened 2021-09-30
Decentralize the configuration as an environment file


t0xic0der:
Decentralize the configuration as an environment file

UPDATE - That environment file should be in the form of a config module, say `config.py`.

DiptoChakrabarty:
I am interested in this issue , just wanted to know by decentralise you mean to put such variables in a .env right?
https://github.com/fedora-infra/mote/blob/7fca9434c1445b1108e532d1f1631baf7eef541a/modules/late.py#L31-L32

t0xic0der:
Exactly.

We have a lot of variables like this one that we'd like to move away to an environment variable file.

subhangi2731:
@t0xic0der @darknao With reference to the discussion in Element, can I please be assigned this issue now, so that I can start working?

PR #189, opened 2021-10-03
Separate the configuration in env files

DiptoChakrabarty:
This is based on issue #180 

A env file has made used to specify the configuration details for the application

python dotenv is used for reading the configs from a .env and used in respective areas of the codebase

sample .env.example has been added to the repository with the previous values 

requirements.txt has been updated

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,3 @@
+BASE_URL="https://apps.fedoraproject.org"
+TOPIC="org.fedoraproject.prod.meetbot.meeting.complete"
+MEETBOT="https://meetbot.fedoraproject.org"
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,3 @@
-BASE_URL="https://apps.fedoraproject.org"
+APP_URL="https://apps.fedoraproject.org"
 TOPIC="org.fedoraproject.prod.meetbot.meeting.complete"
-MEETBOT="https://meetbot.fedoraproject.org"
+BASE_URL="https://meetbot.fedoraproject.org"
--- a/modules/config.py
+++ b/modules/config.py
@@ -3,6 +3,6 @@ import os
 from dotenv import load_dotenv
 load_dotenv()
 
-BASE_URL = os.environ["BASE_URL"]
+APP_URL = os.environ["APP_URL"]
 TOPIC = os.environ["TOPIC"]
-MEETBOT = os.environ["MEETBOT"]
+BASE_URL = os.environ["BASE_URL"]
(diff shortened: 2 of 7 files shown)


mscherer:
So, the code is correct, but I think the configuration naming is not very clear. 

For example "BASE_URL". For other software (like Hugo or Zola), this would be the base url of the current software, eg the URL that will be added in generated link. But here, that's the MEETBOT variable, not BASE_URL.

And BASE_URL is the URL used to query datagrapper, so maybe a more descriptive name could help (for that and MEETBOT ).

DiptoChakrabarty:
> So, the code is correct, but I think the configuration naming is not very clear.
> 
> For example "BASE_URL". For other software (like Hugo or Zola), this would be the base url of the current software, eg the URL that will be added in generated link. But here, that's the MEETBOT variable, not BASE_URL.
> 
> And BASE_URL is the URL used to query datagrapper, so maybe a more descriptive name could help (for that and MEETBOT ).

Shall I rename them to DATAGREPPER_BASE_URL and FEDORA_PROJECT_MEETBOT

mscherer:
I think MEETBOT could be BASE_URL, since that's the meaning of the variable for Hugo, Zola, etc (eg, the base url when creating the website). 

There is no (IMHO) reason to have FEDORA_PROJECT in the name, especially since that's supposed to be reusable by others.

Zlopez:
Wouldn't it be better to use config file (for example https://github.com/fedora-infra/the-new-hotness/blob/master/hotness/config.py) instead of env variables?

DiptoChakrabarty:
Should I convert to dictionary format I have already separated the config to different location

DiptoChakrabarty:
Hey are there any changes that needs to be done here

t0xic0der:
@DiptoChakrabarty, we are looking for adding a config module, say `config.py` instead of a `.env` file. Could you please see if you can rework the code for that?

We already set the label `hacktoberfest-accepted` on your PR so as to ensure that even in the unfortunate circumstance of us not being able to merge your PR, your efforts should not go in vain.

DiptoChakrabarty:
Okay yes Ill try that

DiptoChakrabarty:
You merged another PR even though I started at it earlier and my solution is like you requested it :(

t0xic0der:
@DiptoChakrabarty, this issue ticket fell under one of the quantifiable tasks that our selected intern for Outreachy, @subhangi2731 had to perform. Hence, even if we had your PR come in way before the one she made - we could not be in a position to merge yours.

I did mention of the likeliness of not being able to merge your PR in the second paragraph of https://github.com/fedora-infra/mote/pull/189#issuecomment-955826997. I appreciate the efforts you put into this PR and you have my sincere apologies for not being able to take up your contributions.

Issue #177, opened 2021-09-29
Flash meeting log sections in focus when directed


t0xic0der:
Example - https://css-tricks.com/on-target/

t0xic0der:
@chinmaym07, could you please provide an ETA for this?

chinmaym07:
Hey @t0xic0der . I will take atmost 2 days to complete this issue . Will it work ?

t0xic0der:
2 days is fine. Please limit your PR to what is been requested in the issue ticket.

t0xic0der:
The recent meetings would be inaccessible as the archive was taken some time back. Try opening meetings from the other buttons.

chinmaym07:
I tried all the links which are there in `Last day` & `Last week` tabs but all are giving the same response 404 Not Found . also there are no channels in `explore` ..

t0xic0der:
Please take the instructions provided here https://github.com/t0xic0der/fragment for a spin and let me know if it works.

chinmaym07:
Hey I tried to access this file in drive . But I don't have access to it . I have requested you for access .
https://drive.google.com/file/d/1iOKEgyHMtpnCwsuZa5ufT8gI9-ntuOVc/view?usp=sharing

t0xic0der:
I've shared it with you. :)

chinmaym07:
Hey I have downloaded & extracted the zip file but 404 Not found issue is still there . I tried to log the meetcont fetched using `
fetch_meeting_content` method defined in `modules/call.py`. There is some exception due to which I am getting `(false,'')`.

![image](https://user-images.githubusercontent.com/44670961/135708661-4178fc9c-c056-48e3-921a-1b89673dd5f1.png)

t0xic0der:
Are you sure you extracted the contents in the right location?

chinmaym07:
Yes I think I did ..  
![image](https://user-images.githubusercontent.com/44670961/135753704-d08d3a0a-332a-4bee-baf5-bf5d9462fcb2.png)

t0xic0der:
So the meetings from the recent menu would result in a 404 and that is a
normal behavior as the meeting archive had been taken some time ago. Please
check if you are getting 404 while opening up meetings from either explore
or search menu.

chinmaym07:
Yes that's correct as there are no folder's for the recent meetings due to which it is resulting in 404 which is fine .. therefore I tried to access the last meeting in any folder but that too resulted in 404 .

[...]

PR #190, opened 2021-10-03
Meeting log section should flash when in target

chinmaym07:
Issue: #177 

A new selector is defined in the `fragment.css` file whenever a link is clicked in the logs it has a name attribute with a value `l-`. 
I only needed to select two span elements which are adjacent to the target element.

https://user-images.githubusercontent.com/44670961/135765971-a82efb94-b361-4ca4-8e92-3532c2990f8a.mp4

Signed-off-by: chinmaym07 <b418020@iiit-bh.ac.in>


t0xic0der:
@lilyx13, could you please review this?

static/css3/fragment.css
```diff
@@ -131,3 +131,6 @@ hr {
     color:lime;
 }
 */
+[name^="l-"]:target + .tm, [name^="l-"]:target + .tm + .nk, [name^="l-"]:target + .tm + .nka {
+    background-color: #05ffff;
```
t0xic0der, line 135: Could you please amend the most recent commit to introduce one of the lighter palettes from here https://colorswall.com/palette/24612/ ?

The colors listed in that link belong from the #008080 family but the one that you used does not.

t0xic0der (changes requested):

darknao:
> Hey @t0xic0der Thanks for the review. I really appreciate for your help in setting up the project locally. 1st_place_medal A small query, I see there is one `conflicting file` in this PR should I update it ?

If you can, yes, please rebase your branch and resolve the conflict. If you don't know how, that's ok, I'll do it later today.

PR #201, opened 2021-10-26
created websockets and toast popup for the recently finished meetings

### LINKED ISSUE #167: Use WebSockets and toast popups to inform users of recently finished meetings
Use WebSockets and toast popups to inform users of recently finished meetings

subhangi2731:
Using socket.io to send and receive data across browser and backend server. The function pings the endpoint every 5 seconds to fetch meetings that have been completed in the last 60 seconds and thus informing users for the recently finished meetings

Closes  #167

### DIFF
--- a/src/mote/main.py
+++ b/src/mote/main.py
@@ -20,11 +20,15 @@
     SOFTWARE.
 """
 
+import json
 import re
+import urllib.request as ulrq
 from datetime import datetime
+from threading import Event, Lock
 
 import click
 from flask import Flask, abort, jsonify, render_template, request
+from flask_socketio import SocketIO
 
 from mote.__init__ import __version__
 from mote.modules.call import (
@@ -38,6 +42,12 @@ from mote.modules.late import fetch_recent_meetings
 
 main = Flask(__name__)
 main.config.from_pyfile("config.py")
+socketio = SocketIO(main)
+recognition_pattern = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
+thread = None
+thread_lock = Lock()
+client_count = 0
+exit_thread = Event()
 
 
 @main.get("/fragedpt/")
@@ -123,6 +133,43 @@ def mainpage():
     return render_template("mainpage.html")
 
 
+def fetch_recently_completed_meetings():
+    while not exit_thread.is_set():
+        datagrepper_base_url = "https://apps.fedoraproject.org"
+        topic = "org.fedoraproject.prod.meetbot.meeting.complete"
+        call_period = 60
+        source = "{}/datagrepper/raw?delta={}&topic={}".format(
+            datagrepper_base_url, call_period, topic
+        )
+        parse_object = json.loads(ulrq.urlopen(source).read().decode())
+        meeting_count = parse_object["total"]
+        if meeting_count != 0:
+            meeting_rawlist = parse_object["raw_messages"]
+            socketio.emit("show_toast", meeting_rawlist)
+        exit_thread.wait(60)
+    exit_thread.clear()
+
+
+@socketio.on("connect")
+def trigger_on_connect():
+    global thread
+    global client_count
+    client_count += 1
+    with thread_lock:
+        if thread is None:
+            thread = socketio.start_background_task(fetch_recently_completed_meetings)
+
+
+@socketio.on("disconnect")
+def trigger_on_disconnect():
+    global client_count
+    global thread
+    client_count -= 1
+    if client_count == 0:
+        exit_thread.set()
+        thread = None
+
+
 @click.command()
 @click.option(
     "-p", "--portdata", "portdata", help="Set the port value [0-65536]", default="9696"
@@ -162,4 +209,4 @@ def page_not_found(error):
 
 
 if __name__ == "__main__":
-    mainfunc()
+    socketio.run(mainfunc())
--- a/src/mote/templates/mainpage.html
+++ b/src/mote/templates/mainpage.html
@@ -1,4 +1,3 @@
-
 <!doctype html>
 <!--
 ##########################################################################
@@ -33,6 +32,7 @@
         <link rel="icon" href="{{ url_for('static', filename='imgs/logoteal.png') }}" sizes="16x16">
         <script src="{{ url_for('static', filename='jscn/fragment.js') }}"></script>
         <script src="{{ url_for('static', filename='jscn/easy.qrcode.min.js') }}"></script>
+        <script src="{{ url_for('static', filename='jscn/socket.io.js') }}"></script>
         <title>Fragment</title>
     </head>
     <body class="body bg-light">
@@ -76,12 +76,12 @@
                             <p class="h2 head" style="font-weight: bold;">Fragment</p>
                             <p class="h4 body">Conversation Caretaker</p>
                         </div>
-                        <a style="text-decoration: none; 
+                        <a style="text-decoration: none;
                         "href="#" class="btn btn-light mt-3">
                             <i class="far fa-dot-circle h1 tealfilt"></i><br/>
                             <span style="color: #676767;">About</span>
                         </a>
-                        <a style="text-decoration: none; 
+                        <a style="text-decoration: none;
                         "href="#"
                            class="btn btn-light mt-3"
                            data-bs-toggle="modal"
@@ -101,7 +101,7 @@
                             <i class="far fa-arrow-alt-circle-right h1 tealfilt"></i><br/>
                             <span style="color: #676767;">Explore</span>
                         </a>
-                        <a style="text-decoration: none; 
+                        <a style="text-decoration: none;
                         "href="#"
                            class="btn btn-light mt-3"
                            data-bs-toggle="modal"
@@ -236,7 +236,7 @@
         <div class="modal fade" aria-hidden="true" data-bs-backdrop="static" data-bs-keyboard="true" id="datemode">
             <div class="modal-dialog modal-dialog-centered modal-xl modal-fullscreen-lg-down modal-dialog-scrollable">
                 <div class="modal-content">
-                    <div class="modal-header justify-content-center p-2" style="background-image: linear-gradient(180deg,#eee 0,#ddd); color:#51A2DA;"> 
+                    <div class="modal-header justify-content-center p-2" style="background-image: linear-gradient(180deg,#eee 0,#ddd); color:#51A2DA;">
                         <h6 class="modal-title head fw-bold">Explore meetings</h6>
                     </div>
                     <div class="modal-body">
@@ -394,5 +394,43 @@
                 </div>
             </div>
         </div>
+        <div class="toast-container">
+            <div id="parent" class="position-fixed bottom-0 end-0 p-3" style="z-index: 11">
+            </div>
+        </div>
     </body>
+    <script>
+        var socket = io.connect();
+        function showToast(meeting_json) {
+            meeting_count = meeting_json.length;
+            const timer = ms => new Promise(res => setTimeout(res, ms));
+            async function load() {
+                for (var i = 0; i < meeting_count; i++) {
+                    var toastDiv = $(`
+                        <div class="toast-body">
+                            <h6 id="meeting-name">`+ meeting_json[i]["msg"]["meeting_topic"] +`</h6>
+                            <div class="mt-2 pt-2 border-top">
+                                <a id="toast-btn" type="button" class="btn btn-primary btn-sm" href="`+ meeting_json[i]["msg"]["url"] +`" target="_blank">Open meeting</a>
+                                <button type="button" class="btn btn-secondary btn-sm" data-bs-dismiss="toast">Close</button>
+                            </div>
+                        </div>
+                    `);
+                    var parentDiv=document.getElementById('parent');
+                    var div = document.createElement('div');
+                    div.setAttribute('class','toast');
+                    div.setAttribute('role','alert');
+                    div.setAttribute('aria-live','assertive');
+                    div.setAttribute('aria-atomic','true');
+                    $(div).append(toastDiv);
+                    $(parentDiv).append(div);
+                    var toast = new bootstrap.Toast(div, {delay: 10000});
+                    toast.show();
+                }
+            }
+            load();
+        }
+        socket.on("show_toast", function(meeting_json) {
+            showToast(meeting_json);
+        });
+    </script>
 </html>
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 beautifulsoup4>=4.0.0
 Flask>=2.0.0
+Flask-SocketIO>=5.1.1
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setuptools.setup(
     include_package_data=True,
     zip_safe=False,
     packages=setuptools.find_packages("src"),
-    install_requires=["flask", "beautifulsoup4"],
+    install_requires=["flask", "beautifulsoup4", "flask-socketio"],
     # automatically create console scripts
     entry_points={
         "console_scripts": [
--- a/src/mote/static/jscn/socket.io.js
+++ b/src/mote/static/jscn/socket.io.js
@@ -1,0 +1,4181 @@
+/*!
+ * Socket.IO v4.3.2
+ * (c) 2014-2021 Guillermo Rauch
+ * Released under the MIT License.
+ */
+(function (global, factory) {
+  typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
+  typeof define === 'function' && define.amd ? define(factory) :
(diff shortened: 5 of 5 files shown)


src/mote/templates/mainpage.html
```diff
@@ -394,5 +395,34 @@ <h6 class="modal-title head fw-bold">Recent conversations</h6>
...
+                    </div>
+                </div>
+            </div>
+        </div>
     </body>
+    <script>
+      var socket = io.connect("http://" + document.domain + ":" + location.port)
+      socket.on("connect", function() {
+        socket.emit("trigger function")
+      } )
```
darknao, line 416: If you listen to the 'connect' event, you don't need to send any additional messages, and can remove this block.

src/mote/main.py
```diff
@@ -123,6 +128,26 @@ def mainpage():
...
+    )
+    parse_object = json.loads(ulrq.urlopen(source).read().decode())
+    meeting_count = parse_object["total"]
+    if meeting_count != 0:
+        meeting_rawlist = parse_object["raw_messages"]
+        socketio.emit("show toast", meeting_rawlist)
+
+
+@socketio.on("trigger function")
+def handle_my_custom_event():
```
darknao, line 147: Try to use a meaningful name, like on_connect or something like that.

src/mote/templates/mainpage.html
```diff
@@ -394,5 +395,34 @@ <h6 class="modal-title head fw-bold">Recent conversations</h6>
                 </div>
             </div>
         </div>
+        <div class="toast-container"> 
+            <div class="position-fixed bottom-0 end-0 p-3" style="z-index: 11">
+                <div id="liveToast" class="toast" role="alert" aria-live="assertive" aria-atomic="true">
+                    <div class="toast-body">
+                        A meeting has just ended
```
darknao, line 402: I think we need a little more information in this notification. The meeting name at least.

src/mote/templates/mainpage.html
```diff
@@ -394,5 +395,34 @@ <h6 class="modal-title head fw-bold">Recent conversations</h6>
...
+                            <a id="toast-btn" type="button" class="btn btn-primary btn-sm" href="" target="_blank">Open meeting</a>
+                            <button type="button" class="btn btn-secondary btn-sm" data-bs-dismiss="toast">Close</button>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
     </body>
+    <script>
+      var socket = io.connect("http://" + document.domain + ":" + location.port)
```
darknao, line 413: I don't think you need to pass any arguments if you're using the same server
```suggestion
      var socket = io.connect()
```

darknao (changes requested):
Ok, so this is a little bit suboptimal:
You are starting a new thread every time a client connect.
Each threads will poll datagrepper, and broadcast the result to all connected clients.

Let say we have 5 clients connected, you end up with 5 separates threads that does the exact same thing.
When a meeting end, each thread will send a message to **all** clients (in this case, each client will receive 5 times the same message).
And even worst: those threads are never terminated, even when the client disconnect.
This can cause some serious amplification attack toward the datagrepper service, and we don't want that :)

A better approach would be to only start one thread when at least one client is connected, and stop it when the last client disconnect.

subhangi2731:
@darknao I've updated the PR with all the changes mentioned above. 
The updated PR has the following changes:
* Each time a client connects, the program logic checks if the function for pulling data from Datagrepper endpoint is already running. If it does, it does not run the function again.
* The single running function now broadcasts the toast message to all of the running clients
* When the clients are connected, a counter is kept to maintain a record of the number of clients connected at once. This helps close the running function once all the clients are disconnected. In short, no more API calls once the clients are all disconnected, hence saving resources.
* The function is called when atleast one client connects, and is stopped when the last one disconnects.
* Redundant messages used to handle socket.io events are now removed.
* Function names are made more meaningful
* Toast notification now even has the name of the finished meeting, along with a `Open meeting`  hyperlink and a `Close` button for the toast to close.
* When there are multiple meetings finished in the last minute, the code now shows up a toast for each of them, in a span of `5 seconds` so that the user gets time to react to each toast.

new

subhangi2731:
@darknao @t0xic0der I did all the changes whatever told please check and let me know

src/mote/main.py
```diff
@@ -123,6 +130,42 @@ def mainpage():
...
+
+
+@socketio.on("connect")
+def trigger_on_connect():
+    global client_count
+    global is_running
+    client_count += 1
+    if is_running == False:
+        is_running = True
+        fetch_recently_completed_meetings()
```
darknao, line 157: Few recommendations:
Use thread lock mechanism to avoid race condition.
For the first client connected, this function is blocked until the `fetch_recently_completed_meetings` exit.
Start a background task instead. You can use the socketIO utility function `socketio.start_background_task` for that purpose.

src/mote/main.py
```diff
@@ -123,6 +130,42 @@ def mainpage():
...
+
+
+@socketio.on("disconnect")
+def trigger_on_disconnect():
+    global client_count
+    global is_running
+    client_count -= 1
+    if client_count == 0:
+        is_running = False
```
darknao, line 166: When the last client disconnects, there is a 60 seconds period when `is_running` is False, but the while loop is still running the sleep call.
If another client connects during this time, it starts a new loop, and the previous one keeps running.
You need to make sure the loop is stopped before running a new one.

src/mote/templates/mainpage.html
```diff
@@ -394,5 +395,42 @@ <h6 class="modal-title head fw-bold">Recent conversations</h6>
...
+            for (var i = 0; i < meeting_count; i++) {
+                console.log(i)
+                var toastLink = document.getElementById("toast-btn")
+                var meetName = document.getElementById("meeting-name")
+                meetName.innerHTML = "A meeting has just ended: " + meeting_json[i]["msg"]["meeting_topic"]
+                toastLink.href = meeting_json[i]["msg"]["url"]
+                var toast = new bootstrap.Toast(toastElement)
+                toast.show() 
+                await timer(5000)
+            }
```
darknao, line 428: Instead of doing this one by one, maybe we can stack notifications ?

darknao (changes requested):
You can take inspiration from this example code from socketIO to see how to handle background task in a cleaner way : https://github.com/miguelgrinberg/Flask-SocketIO/blob/main/example/app.py

subhangi2731:
I have implemented stacking of toasts currently, working on implementing background task now.

src/mote/main.py
subhangi2731: Used socketio's background task now, instead of manual triggering through the use of global variables.

subhangi2731:
@darknao I have made all the changes. 
* Socketio's background task feature is now used to start the function
* A threading Event now terminates the function when set to `True` despite any sleep/wait cycle.

src/mote/main.py
```diff
@@ -123,6 +133,43 @@ def mainpage():
...
+        call_period = 60
+        source = "{}/datagrepper/raw?delta={}&topic={}".format(
+            datagrepper_base_url, call_period, topic
+        )
+        parse_object = json.loads(ulrq.urlopen(source).read().decode())
+        meeting_count = parse_object["total"]
+        if meeting_count != 0:
+            meeting_rawlist = parse_object["raw_messages"]
+            socketio.emit("show_toast", meeting_rawlist)
+            exit.wait(60)
```
darknao, line 149:
```suggestion
        exit.wait(60)
```

src/mote/main.py
```diff
@@ -37,7 +42,12 @@
 from mote.modules.late import fetch_recent_meetings
 
 main = Flask(__name__)
+socketio = SocketIO(main)
 recognition_pattern = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
+thread = None
+thread_lock = Lock()
+client_count = 0
+exit = Event()
```
darknao, line 50: `exit` is already used internally by Python, choose another name, like `exit_thread`.

darknao (changes requested):
Ok, that's nice.
I still have only one notification showing after another, instead of all of them at once.
And maybe we should leave them a bit longer ? 5 seconds seems short to me.

subhangi2731:
@darknao 
* I have removed wait time when showing toasts, since that is the reason for toasts being displayed individually and not in a stack.
* I have updating with better naming conventions.

subhangi2731:
@t0xic0der @darknao please review my updated PR thanks!

darknao (approved):

darknao:
Seems good to me :+1: 
I still think the 5 seconds display time is a bit too fast though...

subhangi2731:
@darknao Should I decrease the display time?

darknao:
> @darknao Should I decrease the display time?
I mean increase it a little. Like 8 or 10 seconds maybe?

subhangi2731:
@darknao sure

subhangi2731:
@darknao can you please point me out where do I need to increase the display time as the 5-second time has already been removed, because it led to single toasts being shown instead of in stack

src/mote/templates/mainpage.html
```diff
@@ -394,5 +395,43 @@ <h6 class="modal-title head fw-bold">Recent conversations</h6>
...
+                `)
+                var parentDiv=document.getElementById('parent')
+                var div = document.createElement('div')
+                div.setAttribute('class','toast')
+                div.setAttribute('role','alert')
+                div.setAttribute('aria-live','assertive')
+                div.setAttribute('aria-atomic','true')
+                $(div).append(toastDiv)
+                $(parentDiv).append(div)
+                var toast = new bootstrap.Toast(div)
```
darknao, line 427: One way of increasing the delay (in ms):
```suggestion
                var toast = new bootstrap.Toast(div, {delay: 10000})
```

t0xic0der:
@subhangi2731, please provide us with updates on this if there are any.

subhangi2731:
@t0xic0der @darknao sure please check and let me know

darknao:
Looks good.
Please rebase your branch and solve conflicts, then it should be good to merge.

t0xic0der:
@darknao, could you please provide me with the steps you performed to review the PR - function-wise?

subhangi2731:
@darknao okay I will do it

darknao:
> @darknao, could you please provide me with the steps you performed to review the PR - function-wise?

Old fashioned manual testing using logging and playing with the call_period of datagrepper for python side, and dev console for the js side.

subhangi2731:
@darknao This has no conflicts, so what else am I expected to do here as I resolved everything and all the checks passed

t0xic0der:
@darknao, aren't this and https://github.com/fedora-infra/mote/pull/218 ending up doing the same thing? I'm curious.

darknao:
#218 is based on this PR (they share the same commit history)
This one is adding the notification part using websocket and datagrepper for fetching recent meetings.
The other one replace the datagrepper part with fedora-messaging.

t0xic0der:
Makes sense. I'll give this PR another look once I have an merge approval from your end.

src/mote/main.py
```diff
@@ -20,11 +20,16 @@
     SOFTWARE.
 """
 
+import json
 import re
+import time
```
t0xic0der, line 25: I don't see the `time` library being used here. Please remove its import.

src/mote/templates/mainpage.html
```diff
@@ -394,5 +395,43 @@ <h6 class="modal-title head fw-bold">Recent conversations</h6>
...
+                $(parentDiv).append(div)
+                var toast = new bootstrap.Toast(div, {delay: 10000})
+                toast.show()
+            }
+        }
+        load()
+      }
+      socket.on("show_toast", function(meeting_json) {
+        showToast(meeting_json)
+      } )
```
t0xic0der, line 435: This is nitpicking but can you please make the indentation uniform for the JavaScript block?

src/mote/templates/mainpage.html
```diff
@@ -33,6 +33,7 @@
         <link rel="icon" href="{{ url_for('static', filename='imgs/logoteal.png') }}" sizes="16x16">
         <script src="{{ url_for('static', filename='jscn/fragment.js') }}"></script>
         <script src="{{ url_for('static', filename='jscn/easy.qrcode.min.js') }}"></script>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/socket.io/4.3.2/socket.io.js"></script>
```
t0xic0der, line 36: Please download and store the JavaScript library in `src/mote/static/jscn/` and reference it from there - just like we're doing for other such JavaScript dependencies.

t0xic0der (changes requested):
Some more suggested changes.

t0xic0der:
I added https://github.com/fedora-infra/mote/pull/201/commits/916b69a29ba76fb8951e6fdf88496da6ee49067c. Please be sure to pull before making any changes.

subhangi2731:
@t0xic0der sure I will do this

subhangi2731:
@t0xic0der I resolved the indentation and removed time import but where is the js source file for socket io I can't find it in the website can you please tell

src/mote/templates/mainpage.html
t0xic0der, line 36: As a reply to https://github.com/fedora-infra/mote/pull/201#issuecomment-982374808, the source is present in this [line](https://github.com/fedora-infra/mote/pull/201/files#diff-4c11e602a99bd1a8ca037ca1be41a6574c9c32ce26b2a7bee39fdd0c46fec471R36).

subhangi2731: @t0xic0der yes yes I got it thanks

subhangi2731:
@t0xic0der @darknao I rebased the branch resolved all the conflicts that came in the way (mainpage.html and main.py) and did all the changes as said and added the dependency please check and let me know

darknao:
looks good to me.
I don't see the commit from @t0xic0der anymore, but it seems to be included in your commit now, so I guess it's ok.
There is still a merge commit in the history, which mean a rebase is needed.

subhangi2731:
@darknao yes due to the conflicts it created an issue but I downloaded the dependency and included it. Ok I will just do the rebase thing

subhangi2731:
@darknao @t0xic0der  please have a look now

darknao:
Yep, that's fine for me :+1:

src/mote/templates/mainpage.html
```diff
@@ -394,5 +395,43 @@ <h6 class="modal-title head fw-bold">Recent conversations</h6>
...
+                $(parentDiv).append(div)
+                var toast = new bootstrap.Toast(div, {delay: 10000})
+                toast.show()
+            }
+        }
+        load()
+        }
+        socket.on("show_toast", function(meeting_json) {
+            showToast(meeting_json)
+        }     )
```
t0xic0der, line 435:
```suggestion
        var socket = io.connect();
        function showToast(meeting_json) {
            meeting_count = meeting_json.length;
            const timer = ms => new Promise(res => setTimeout(res, ms));
            async function load() {
		        for (var i = 0; i < meeting_count; i++) {
		            var toastDiv = $(`
		                <div class="toast-body">
		                    <h6 id="meeting-name">`+ meeting_json[i]["msg"]["meeting_topic"] +`</h6>
		                    <div class="mt-2 pt-2 border-top">
		                        <a id="toast-btn" type="button" class="btn btn-primary btn-sm" href="`+ meeting_json[i]["msg"]["url"] +`" target="_blank">Open meeting</a>
		                        <button type="button" class="btn btn-secondary btn-sm" data-bs-dismiss="toast">Close</button>
		                    </div>
		                </div>
		            `);
		            var parentDiv=document.getElementById('parent');
		            var div = document.createElement('div');
		            div.setAttribute('class','toast');
		            div.setAttribute('role','alert');
		            div.setAttribute('aria-live','assertive');
		            div.setAttribute('aria-atomic','true');
		            $(div).append(toastDiv);
		            $(parentDiv).append(div);
		            var toast = new bootstrap.Toast(div, {delay: 10000});
		            toast.show();
		        }
        	}
        	load();
        }
        socket.on("show_toast", function(meeting_json) {
            showToast(meeting_json);
        });
```

t0xic0der (changes requested):
Oh I so passionately hate the ES6 inline functions!

Anyway, here are some suggested changes.

subhangi2731:
@t0xic0der done

t0xic0der:
@subhangi2731, I noticed that you amended the most recent commit https://github.com/fedora-infra/mote/commit/9443aa618495395abcc4680214bc5107ac7adefb to accommodate the changes that I suggested on https://github.com/fedora-infra/mote/pull/201#pullrequestreview-819783983 even though the changes are disjointed from the commit message. Please try to keep like changes together and if needed, you can always create a new commit to address unlike changes rather than pitting together the unlike ones.

t0xic0der (approved):
Sigh.

LGTM.

subhangi2731:
@t0xic0der sure I will take care of this from the next time. The problem happened as there were many conflicts I had to rebase the branch two times to fix the commits. I thought to include all commit messages in one but I will create a new commit the next time for the particular changes sure

PR #202, opened 2021-10-27
Created url shortening application

tuba-mouqeem:
Please review and suggest, how can I make it better?

### DIFF
--- a/src/mote/url_shortner.py
+++ b/src/mote/url_shortner.py
@@ -1,18 +1,20 @@
-from flask import Flask, request, render_template
+from flask import Flask, render_template, request
 from flask_sqlalchemy import SQLAlchemy
 
 app = Flask(__name__)
-app.config['SQLALCHEMY_DATABASE_URI'] = "sqlite:///url.db"
-app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
+app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///url.db"
+app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
 db = SQLAlchemy(app)
+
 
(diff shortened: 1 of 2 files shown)


darknao:
I'm not sure how to review this...
Can you explain what you have in mind? or provide a few examples on how to use this function?
I see a sqlite database, but you never use it to store anything. Did I miss something? Is this code working?

src/mote/url_shortner.py
```diff
@@ -0,0 +1,62 @@
...
+        else:
+            self.Storing_UrlId[longUrl] = self.id
+            shorten_url = self.convereter_func(self.id)
+            # Counter increment
+            self.id = self.id + 1
+
+        return "meetbot.fedoraproject.org/" + shorten_url
+
+    # Converting b10 to b62(Alphanumeric)
+    def convereter_func(self, id):
```
Zlopez, line 43:
```suggestion
    def converter_func(self, id):
```

darknao (changes requested):
There are some good ideas here, but the implementation is far from complete.
Possible improvements are:
- Create 2 functions, one to convert a long url to a short one, and another to do the reverse (short to long url)
- The first function should check if the url already exist in db, and if not, create a new ShortUrl instance, generate the short url, save it into the db, then return the short url.
- The second one should query the db for the desired short url, and return the corresponding long one.
- Combine the 2 classes into one: move `converter_func` to the ShortUrl class, and rename it to something more meaningful like `get_short_url`. This function should use the object `id` (available as `self.id`)
- Create a route endpoint that take a short url in parameter, and redirect to the matching long url (something like `/<short_url>` or `/s/<short_url>`)

t0xic0der:
Closing this as stale

PR #204, opened 2021-11-01
Create separate config file for frequently used configuration

### LINKED ISSUE #180: Decentralize the configuration as an environment file

subhangi2731:
A new `config.py` has been created to store the configuration that is commonly used across the project.

Closes #180

### DIFF
--- a/src/mote/config.py
+++ b/src/mote/config.py
@@ -1,0 +1,4 @@
+MEETING_DIR = "/srv/web/meetbot"
+MEETBOT_URL = "https://meetbot-raw.fedoraproject.org"
+DATAGREPPER_BASE_URL = "https://apps.fedoraproject.org"
+RECOGNIITION_PATTERN = r"(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
--- a/src/mote/main.py
+++ b/src/mote/main.py
@@ -37,7 +37,7 @@ from mote.modules.find import find_meetings_by_substring
 from mote.modules.late import fetch_recent_meetings
 
 main = Flask(__name__)
-recognition_pattern = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
+main.config.from_pyfile("config.py")
 
 
 @main.get("/fragedpt/")
@@ -90,8 +90,8 @@ def fragedpt():
 @main.get("/<channame>/<cldrdate>/<path:meetname>")
 def statfile(channame, cldrdate, meetname):
     meetname = meetname.replace(".log.html", "").replace(".html", "")
-    meeting_title = re.search(recognition_pattern, meetname)
-    meetpath = "/srv/web/meetbot" + request.path
+    meeting_title = re.search(main.config["RECOGNIITION_PATTERN"], meetname)
+    meetpath = main.config["MEETING_DIR"] + request.path
     formatted_timestamp = datetime.strptime(cldrdate, "%Y-%m-%d")
     cldrdate = "{:%B %d, %Y}".format(formatted_timestamp)
     print(meetpath)
--- a/src/mote/modules/call.py
+++ b/src/mote/modules/call.py
@@ -26,16 +26,15 @@ import urllib.parse as ulpr
 from datetime import datetime
 
 import bs4 as btsp
-
-recognition_pattern = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
+from flask import current_app as app
 
 
 def fetch_channel_dict():
     try:
         channel_dict = {}
-        chanlist = os.listdir("/srv/web/meetbot")
+        chanlist = os.listdir(f"{app.config['MEETING_DIR']}")
         for channel in chanlist:
-            channel_dict[channel] = "https://meetbot-raw.fedoraproject.org/%s" % channel
+            channel_dict[channel] = f"{app.config['MEETBOT_URL']}/{channel}"
         return True, channel_dict
     except Exception as expt:
         return False, {"exception": str(expt)}
@@ -44,12 +43,9 @@ def fetch_channel_dict():
 def fetch_datetxt_dict(channel: str):
     try:
         datetxt_dict = {}
-        datelist = os.listdir("/srv/web/meetbot/%s" % channel)
+        datelist = os.listdir(f"{app.config['MEETING_DIR']}/{channel}")
         for datetxt in datelist:
-            datetxt_dict[datetxt] = "https://meetbot-raw.fedoraproject.org/%s/%s" % (
-                channel,
-                datetxt,
-            )
+            datetxt_dict[datetxt] = f"{app.config['MEETBOT_URL']}/{channel}/{datetxt}"
         return True, datetxt_dict
     except Exception as expt:
         return False, {"exception": str(expt)}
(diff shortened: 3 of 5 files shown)


subhangi2731:
@darknao @t0xic0der I have included the config file in `src/mote` directory. Let me know if it's preferable to be kept in a separate `configs` folder, maybe. Also, the config file uses class variables, that promote better readability, structured access and needs minimal changes in the future when new configs are to be added.

mscherer:
I feel there is a overlap  with #189 (linked in #180)

subhangi2731:
@mscherer I am assigned to work on this issue as a part of the `Outreachy` program.

darknao:
There is already a config facility included and used by Flask and all its modules. Should we use that instead?
see https://flask.palletsprojects.com/en/2.0.x/config/ for more details.

t0xic0der (changes requested):
@darknao, yep. It makes a lot more sense to use something that is "cooked into" the framework that we are making use of. 

@subhangi2731, could you please follow up with @darknao's suggestion here and make changes to your code?

@mscherer, yep. We shifted goalposts there so instead of the use of an environment variable file, we would now want to use a config module.

subhangi2731:
@t0xic0der @darknao okay will do the config module. Going ahead with @darknao suggestions

subhangi2731:
@darknao @t0xic0der I have updated the PR with the built-in configuration handler by Flask. Please review.

darknao (changes requested):
Looks good, just one place needed to fix.

src/mote/modules/find.py
```diff
@@ -26,7 +26,9 @@
 import urllib.parse as ulpr
 from datetime import datetime
 
-directory_path = os.path.dirname("/srv/web/meetbot/")
+from flask import current_app as app
+
+directory_path = os.path.dirname("/srv/web/meetbot")
```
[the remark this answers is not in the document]
subhangi2731: The two lines have been moved into the function. Otherwise, the value is accessed before the application is initialized. This causes the flask to give out an error.

src/mote/modules/find.py
```diff
@@ -26,24 +26,25 @@
...
-recognition_pattern = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
+from flask import current_app as app
 
 
 def find_meetings_by_substring(search_string: str):
     """
     Return list of meetings returned from search
     """
+    directory_path = os.path.dirname(app.config["MEETING_DIR"])
+    recognition_pattern = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
```
darknao, line 37: You don't need to move that one.
Note that I see this same variable at three different places already. Maybe we can make it a configuration item too?

subhangi2731:
@darknao I have added a separate configuration for `recognition_pattern` as well, and it was needed to be moved into the functions, because otherwise the value would be accessed before the app initializes, which gives out an error.

subhangi2731:
@t0xic0der @darknao please review my updated PR

src/mote/main.py
```diff
@@ -37,7 +37,8 @@
 from mote.modules.late import fetch_recent_meetings
 
 main = Flask(__name__)
-recognition_pattern = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
+main.config.from_pyfile("config.py")
+recognition_pattern = main.config["RECOGNIITION_PATTERN"]
```
darknao, line 41: You don't need this line at this point, you can simply use `main.config["RECOGNIITION_PATTERN"]` wherever it's needed directly

src/mote/modules/call.py
```diff
@@ -57,18 +56,19 @@ def fetch_datetxt_dict(channel: str):
 
 def fetch_meeting_dict(channel: str, datetxt: str):
     try:
+        recognition_pattern = app.config["RECOGNIITION_PATTERN"]
```
darknao, line 59: same here

src/mote/modules/find.py
```diff
@@ -26,24 +26,25 @@
...
-recognition_pattern = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
+from flask import current_app as app
 
 
 def find_meetings_by_substring(search_string: str):
     """
     Return list of meetings returned from search
     """
+    directory_path = os.path.dirname(app.config["MEETING_DIR"])
+    recognition_pattern = app.config["RECOGNIITION_PATTERN"]
```
darknao, line 37: and here

src/mote/modules/late.py
```diff
@@ -25,12 +25,14 @@
...
 
+from flask import current_app as app
+
 seconds_delta = 86400
 
 
 def fetch_recent_meetings(days):
     try:
-        datagrepper_base_url = "https://apps.fedoraproject.org"
+        datagrepper_base_url = app.config["DATAGREPPER_BASE_URL"]
```
darknao, line 35: same remark as `recognition_pattern`: you can remove this line and use `app.config["DATAGREPPER_BASE_URL"]` directly

subhangi2731:
okay got it I will do this @darknao

subhangi2731:
@darknao please check and review and let me know

src/mote/config.py
```diff
@@ -0,0 +1,4 @@
+MEETING_DIR = "/srv/web/meetbot"
+MEETBOT_URL = "https://meetbot-raw.fedoraproject.org"
+DATAGREPPER_BASE_URL = "https://apps.fedoraproject.org"
+RECOGNIITION_PATTERN = "(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
```
darknao, line 4: You should always use raw string for rexexp
```suggestion
RECOGNIITION_PATTERN = r"(.*)[\-\.]([0-9]{4}-[0-9]{2}-[0-9]{2})-([0-9]{2}\.[0-9]{2})"
```

src/mote/modules/call.py
```diff
@@ -26,16 +26,15 @@
...
 
 
 def fetch_channel_dict():
     try:
         channel_dict = {}
-        chanlist = os.listdir("/srv/web/meetbot")
+        chanlist = os.listdir(app.config["MEETING_DIR"])
         for channel in chanlist:
-            channel_dict[channel] = "https://meetbot-raw.fedoraproject.org/%s" % channel
+            channel_dict[channel] = app.config["MEETBOT_URL"] + "/%s" % channel
```
darknao, line 37: Instead of combining string concatenation & (old) %-format, you should use f-string everywhere. It improve readability, and is actually a bit faster. 
Ex.:
```suggestion
            channel_dict[channel] = f"{app.config['MEETBOT_URL']}/{channel}"
```

darknao (commented):
Looks good to me.
I've added a couple of remarks, but that's only because I'm picky.

subhangi2731:
@darknao sure I will do this and update now

subhangi2731:
@darknao I have done the changes as suggested by you please check and let me know

src/mote/modules/call.py
```diff
@@ -44,12 +43,9 @@ def fetch_channel_dict():
...
     try:
         datetxt_dict = {}
-        datelist = os.listdir("/srv/web/meetbot/%s" % channel)
+        datelist = f"{os.listdir(app.config['MEETING_DIR'])}/{channel}"
         for datetxt in datelist:
-            datetxt_dict[datetxt] = "https://meetbot-raw.fedoraproject.org/%s/%s" % (
-                channel,
-                datetxt,
-            )
+            datetxt_dict[datetxt] = f"{app.config['MEETBOT_URL']}/{channel, datetxt}"
```
darknao, line 48: Make sure the original formatting is respected.
It should be:
```suggestion
            datetxt_dict[datetxt] = f"{app.config['MEETBOT_URL']}/{channel}/{datetxt}"
```

subhangi2731:
@darknao yes okay sure

subhangi2731:
@darknao is it okay now?

src/mote/modules/find.py
```diff
@@ -68,13 +68,15 @@ def find_meetings_by_substring(search_string: str):
...
                                         channel_name,
                                         meeting_date,
                                         meeting_log_filename,
                                     ),
-                                    "summary": "https://meetbot-raw.fedoraproject.org/%s/%s/%s"
+                                    "summary": app.config["MEETBOT_URL"]
+                                    + "/%s/%s/%s"
                                     % (
                                         channel_name,
                                         meeting_date,
```
darknao, line 82: Can you use f-string here too?

subhangi2731:
@darknao sure

src/mote/modules/find.py
subhangi2731: @darknao done

subhangi2731:
@darknao can you please check and let me know if it is done?

src/mote/modules/call.py
```diff
@@ -44,12 +43,9 @@ def fetch_channel_dict():
 def fetch_datetxt_dict(channel: str):
     try:
         datetxt_dict = {}
-        datelist = os.listdir("/srv/web/meetbot/%s" % channel)
+        datelist = f"{os.listdir(app.config['MEETING_DIR'])}/{channel}"
```
darknao, line 46: This is not correct. `datelist` is supposed to be a list, now it's a string representing a list.
The next for-loop will not produce the expected result.
```suggestion
        datelist = os.listdir(f"{app.config['MEETING_DIR']}/{channel}")
```
You have the same mistake at different places in this file.

subhangi2731: @darknao ok I will resolve this

subhangi2731:
@darknao I resolved can you please confirm?

darknao (approved):
Looks good :+1:

t0xic0der (approved):
LGTM.

Merging.

PR #200, opened 2021-10-18
Format the date strings in the listing 

subhangi2731:
In this issue, I have formatted the date strings and grouped according to year and month. I have also attached a screenshot where now the newly formatted date time looks like this. 

Fixes issue #188 
Screenshot from 2021-10-18 23-28-18

PR #205, opened 2021-11-01
remove unused ctime import

subhangi2731:
With reference to PR #200 `ctime` is no longer being used in `late.py` and hence is redundant and should be removed.

@darknao @t0xic0der


darknao (approved):
LGTM

PR #218, opened 2021-11-24
Implement Fedora messaging notifications

darknao:
I've took the work from @subhangi2731 and adapt it a little to work with fedora-messaging.
I've switched the docker image from alpine to fedora-minimal as it is way faster to build and don't need to compile most of the python packages.

Notifications have been tweaked a bit to keep them minimal and less intrusive. They stay on screen until dismissed, as it's very unlikely the user happen to be actively monitoring the page in the 5 seconds time frame a meeting just ended.
image
For now, the meeting link is the one provided in the fedora-messaging bus, and redirect to meetbot.fp-o.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,15 @@
-FROM python:3.8-alpine
-COPY . /opt/app
-LABEL maintainer "Anurag Phadnis <phadnis.anurag@gmail.com>"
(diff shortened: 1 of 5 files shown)


t0xic0der (approved):
LGTM code-wise. I didn't take it for a spin though.

PR #219, opened 2021-11-29
add basic unit testing

darknao:
This adds some tests units which can be run with `tox`.
The coverage is not 100% yet, but it's a start.

Use `tox` for running tests, and `tox -e format` for black+isort.


t0xic0der:
Do we wait on some progress to be made here for the coverage to reach 100% or can it be reviewed now?

darknao:
Current coverage is 93.26%.

There are currently 2 tests failing and preventing us to get full coverage.
Those are about raw txt meeting logs, like <mote_url>/fedora-meeting/2020-06-10/fedora_iot_working_group_meeting.2020-06-10-14.10.txt or <mote_url>/fedora-meeting/2020-06-10/fedora_iot_working_group_meeting.2020-06-10-14.10.log.txt. Both return 404.
And the following code is never executed :
https://github.com/fedora-infra/mote/blob/f97e5b92ab93d43cb9c9c71540465e2ef40ad7b4/src/mote/main.py#L117
I guess something is missing in this function `statfile` to be able to return those kind of logs.
While I could probably do it myself in this PR, and if this is still a required feature, I think we should instead open a dedicated issue to fix it.

Beside that, I think I just need to add a couple of tests for the last PRs that just got merged, and we should be all set.

t0xic0der:
If it helps, I and @subhangi2731 added code quality checks and setup @dependabot alerts for the repo. Please feel free to reorient the tests according to those.

darknao:
Coverage is not 100% yet, but it should be enough to catch most of the issues with dependencies (I'm looking at you cachelib)

PR #231, opened 2021-12-14
add calendar view

subhangi2731:
The calendar view is added with Vanilla plugin and add event.

### DIFF
--- a/src/mote/static/css3/calendar.css
+++ b/src/mote/static/css3/calendar.css
@@ -1,0 +1,128 @@
+/* body {
+    display: flex;
+    margin-top: 50px;
+    justify-content: center;
+    background-color: #FFFCFF;
+  } */
+  button {
+    width: 75px;
+    cursor: pointer;
+    box-shadow: 0px 0px 2px gray;
+    border: none;
+    outline: none;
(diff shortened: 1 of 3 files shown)


subhangi2731:
Screenshot from 2021-12-15 09-48-58

t0xic0der (changes requested):
Please correct me if I'm mistaken but I do remember to have discussed the following three points with you during the yesterday's weekly Websites and Apps team meeting.
1. We agreed upon the use of a dedicated calendar JS library and not a vanilla one to simplify our dealings with the calendar element and avoid any issues there may arise due to a ground-up implementation.
2. We talked about fetching the meetings from the endpoint where the recent meetings are acquired and place them on the calendar for dates extending upto one month. 
3. As Mote is supposed to be a simple application for accessing information about meetings, having any interaction with the calendar to add events DO NOT serve any purpose.

I do not think that this pull request satisfies any of the aforementioned points.

subhangi2731:
@t0xic0der sure this pr can be closed then I will send a different pr with full calendar plugin

darknao:
I don't think we need this anymore.

PR #232, opened 2021-12-21
add calendar and fetch recent meetings in a month

### LINKED ISSUE #228: Rework the landing page
With the significant amount of inputs we accrued from @mairin and @misc, I have a plan that can help rework the landing page in a way that it would end up feeling a lot usable and functional. 

image

This illustration is not pixel-perfect and is representative of the relative positions of the elements that we would want to have on the landing page.

I guess some more context would help, so here's some more context ;P

1. The recent section would have a calendar display restricted to only on the month view as going any further, would defeat the purpose of calling it "recent".
2. The channels section would have a list of channels sorted by activity with a textual/graphical activity view at the right side of it. We'll have a button, that would lead to a modal having all the channels.
3. When the width is not enough for both of them to stick aside each other, these sections will stack - the recent section would be on the top while the channels section would be at the bottom.
4. We will move out the search bar from the lookup modal to the page with suggestive search mechanism that lists the top 5 results and then, provides an option to view the remaining results in a modal if the count is more than 5.
5. With so much activity on the landing page, we will have to get rid of the textual information that is present right now - but maybe we will move that to the about modal.

Opinions?

subhangi2731:
Added the calendar `fullcalendar.io` plugin and fetched the data of all the recent meetings in the month grid view. When the calendar page is loaded the function is called and rendered to fetch recent meetings. However, will display the recent meetings in the calendar also as events.

closes #228


subhangi2731:
Screenshot from 2021-12-23 21-19-25
Screenshot from 2021-12-23 21-19-16

subhangi2731:
@darknao @t0xic0der Please review and let me know. I have displayed the meetings of this current month and the meeting title redirects to the meeting details page. For the previous month, I will display the meetings by tomorrow. Also please let me know how the UI looks

t0xic0der (changes requested):
Please add `fedora_messaging` to the `requirements.txt`

t0xic0der (changes requested):
Please widen the navbar controls to occupy the width taken up by the container that follows below.

For instance, the following 

image

Should look like the following

image

t0xic0der (changes requested):
Please remove or reduce the margin at the top of the container.

For instance, the following

image

Should look, more or less, like the following 

image

Prefer the use of the inbuilt margin classes like `mt-$(size)` over styles like `margin-top: $(size)%` as the latter change when the width of the viewport is varied while the former doesn't.

t0xic0der (changes requested):
Please remove the arrow buttons on the top right corner of the calendar as the recent view is only supposed to show the meetings which took place in this month. Also, consider adding a week view like we discussed before to better filter the meetings.

@mairin, one question. Should we keep the month view as a default or the week view when it comes to the recent meetings page?

image

t0xic0der (changes requested):
There's a significant amount of delay from the time when the page is rendered to the time when the calendar along with the list of meetings appear.

So when I load up the page, I get to see the following for a couple of seconds.

Screenshot from 2021-12-24 09-04-02

Before the following comes

Screenshot from 2021-12-24 08-57-24

Now, two-three seconds is absolutely not the end of the world but I am on a fast internet connection and assuming the similar conditions for the entire community using this application would be very wrong on my part.

Hence I would ask you to do two things.

1. Render the empty calendar first on the page, before even reaching out to Datagrepper to get a list of recent meetings. 
2. Show an indeterminate spinner on the page somewhere to let the users know that the list of recent meetings are being fetched and the calendar would be soon populated with those.
   Take for instance, https://github.com/fedora-infra/mote/blob/main/src/mote/static/jscn/fragment.js#L103-L105 of the `populate_channel_list()` function has an indeterminate spinner rendered by default when the modal is loaded up and it is soon replaced/removed when the contents have arrived.

t0xic0der (changes requested):
Please add another link on the navbar and keep it as active to mark that clicking on it would lead to the current page.

For example, the link may be called `Recent meetings` and it can be marked as `active` (by using the said class). It can act as both a title for the current page and an indicator that the landing page would lead to the page which shows recent meetings.

Here's an indicative example.

image

t0xic0der (changes requested):
Please stick with the design that @mairin suggested here https://github.com/fedora-infra/mote/issues/228#issuecomment-985007664. I am attaching the same image here for a quick reference.

image

fullcalendar.io is very extensible and customizable when it comes to styling to I do not think it would be difficult to closely adhere to the design suggestions that she made.

t0xic0der (changes requested):
Hmm. The `today` button is inactive. Any plans for the same?

image

subhangi2731:
@t0xic0der I have done the changes as said. I was thinking can we keep the arrow buttons to navigate between the weeks of the current month and when we are on week view and go to previous weeks the `today` button gets activated which on clicking quickly helps navigate to the current date. Please let me know your opinions also the design suggested by @Mairin contains Fedora Bootstrap implementation for that tab icons and for these npm installation is required. Let me know your suggestions

subhangi2731:
Screenshot from 2021-12-25 13-20-45
Screenshot from 2021-12-25 13-20-31

t0xic0der (changes requested):
The suggestion of the links on the navbar to occupy 100% of the width seems to have been misinterpreted.

It looks like this now.

image

Please take a "closer" look at my review and the screenshots I have attached at https://github.com/fedora-infra/mote/pull/232#pullrequestreview-839714652. The code in the expected form screenshot give enough information on what needs to be done.

t0xic0der (changes requested):
> Please let me know your opinions also the design suggested by @mairin contains Fedora Bootstrap implementation for that tab icons and for these npm installation is required. Let me know your suggestions


There's no such need to do anything of that sort. Please make use of custom CSS classes to make the frontend resemble with the design that she proposed. Fedora Bootstrap is based on Bootstrap 4.3.1 and we would not want to make use of it as Bootstrap 5 (the one that we currently use) is not backwards-compatible with Bootstrap 4.x.

src/mote/templates/mainpage.html
```diff
@@ -33,11 +33,13 @@
         <script src="{{ url_for('static', filename='jscn/fragment.js') }}"></script>
         <script src="{{ url_for('static', filename='jscn/easy.qrcode.min.js') }}"></script>
         <script src="{{ url_for('static', filename='jscn/socket.io.js') }}"></script>
+        <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/fullcalendar@5.10.1/main.css">
+    <script src="https://cdn.jsdelivr.net/npm/fullcalendar@5.10.1/main.js"></script>
```
t0xic0der, line 37: Please download them into the `/static/` subdirectory and load them up from there.

Also, rename `main.css` and `main.js` aptly to be representative of what they do - say, call them `fullcalendar.css` and `fullcalendar.js` (or something like that).

src/mote/templates/mainpage.html
```diff
@@ -72,12 +79,12 @@
             </div>
         </nav>
         <div class="container-lg">
-            <div class="card text-center" style="margin-top: 10%;">
+            <div class="card text-center" style="margin-top: 0%;">
```
t0xic0der, line 82: I do remember mentioning the use of prebuilt Bootstrap margin classes over hardcoded CSS margin statements here https://github.com/fedora-infra/mote/pull/232#pullrequestreview-839715765. Please follow the same and reduce the margin to the `mt-1` scale.

t0xic0der, line 82:
```suggestion
            <div class="card text-center mt-1">
```

src/mote/templates/mainpage.html
```diff
@@ -72,12 +79,12 @@
...
+                <div id="loader"></div>
                 <div class="card-body">
```
t0xic0der, line 84:
```suggestion
                <div class="card-body pt-0">
```

It looks like this now.

image

The padding at the top of the calendar header is not needed. We'll remove it and it would look like the following.

image

t0xic0der (changes requested):
Here are some more changes that I would want to you to introduce.

t0xic0der (changes requested):
Please take inspiration from https://calendar.google.com/calendar/u/0/r/week and https://calendar.google.com/calendar/u/0/r/month to make sure that your calendar views occupy upto 100% of the screen height and NOT more.

Having a scrollbar in the calendar views is not only intrusive but it also does not look very elegant in my honest opinion.

t0xic0der (changes requested):
Please relocate the spinner to be placed right in front of the active "Recent meetings" link in the navbar and then change the link name to "Loading meetings..." until the time the meetings load up. Once they finish loading up, rename it back to "Recent meetings".

subhangi2731:
@t0xic0der sure I am doing the changes

t0xic0der:
In the 1:1 that we had on 28th Dec '21, this is a gist of changes that we would want to make happen to the current PR.

image

subhangi2731:
Screenshot from 2022-01-02 21-04-22

subhangi2731:
@t0xic0der I did all the changes as discussed in the meeting. I made the week view as default and made the option to switch over to the month view for getting recent meetings. I also disabled the arrow buttons for the month view so that it will show the meetings for the current month only and also changed the colors. Also reduced the extra space vertically so that it can come in one scroll. Changing the format to day name and date number one above the other is not possible and is not customizable through any given methods in the docs. Also to disable arrow buttons, I replaced default month and week switching buttons with the custom buttons which changes the view now with the built-in methods and use vanilla JavaScript to disable the arrow buttons when on month view.

subhangi2731:
@t0xic0der please review and let me know.

subhangi2731:
Screenshot from 2022-01-05 11-43-13

subhangi2731:
@t0xic0der @darknao please review my pr and let me know. I have done all the changes as told.

subhangi2731:
Screenshot from 2022-01-05 13-09-19

subhangi2731:
@t0xic0der @darknao I did all the changes please check

subhangi2731:
@t0xic0der @darknao The black lines, I have removed using bootstrap theming so it doesn't raise any incompatibility issues with currently used bootstrap 5. I used custom CSS to style the classes. Grids are now visible in between the days and the calendar adjusts its size according to the screen height

t0xic0der (approved):
Awesome work. Merging!

PR #234, opened 2022-01-05
implementation of search bar to show the list of meetings in dropdown

subhangi2731:
@t0xic0der @darknao  I spent a lot of time figuring out the way to display the list of meetings without interfering with the calendar and came up with the built-in bootstrap's dropdown solution. I have implemented fetching the meetings and displaying them as shown in the ss below, and will implement the show more button after this.

### DIFF
--- a/src/mote/templates/mainpage.html
+++ b/src/mote/templates/mainpage.html
@@ -38,40 +38,56 @@
         <title>Meetbot Logs</title>
     </head>
     <body class="body bg-light">
-        <nav class="navbar sticky-top navbar-expand-lg navbar-light p-1" style="background-image: linear-gradient(180deg,#eee 0,#ddd);">
-            <div class="container-lg"> 
-                <a class="navbar-brand" href="#">
-                    <img class="logo" alt="MEETBOT LOGS" src="{{ url_for('static', filename='imgs/Fedora_logo.png') }}" style="width: 100%; height: auto;" width="200" height="40" />
+        <nav class="navbar navbar-expand-lg navbar-light bg-light"
+            style="background-image: linear-gradient(180deg,#eee 0,#ddd);">
+            <div class="container-fluid">
+                <a class="navbar-brand" href="#" style="padding-top: 0px; padding-bottom: 0px;">
+                    <img class="logo" alt="MEETBOT LOGS" src="{{ url_for('static', filename='imgs/Fedora_logo.png') }}"
+                        style="width: 80%; height: auto; padding-top: 0%;" width="200" height="40" />
                 </a>
-                <button class="navbar-toggler"
-                        type="button"
-                        data-bs-toggle="collapse"
-                        data-bs-target="#navbarSupportedContent"
-                        aria-controls="navbarSupportedContent"
-                        aria-expanded="false"
+                <button class="navbar-toggler" type="button" data-bs-toggle="collapse"
+                        data-bs-target="#navbarSupportedContent" aria-controls="navbarSupportedContent" aria-expanded="false"
                         aria-label="Toggle navigation">
-                    <span class="navbar-toggler-icon"></span>
-                </button>
-                <div class="navbar-collapse collapse" id="navbarSupportedContent">
-                    <ul class="navbar-nav ms-auto">
-                        <li class="nav-item" id="loader">
+                        <span class="navbar-toggler-icon"></span>
+                    </button>
+                <div class="collapse navbar-collapse" id="navbarSupportedContent">  
+                    <ul class="navbar-nav me-auto mb-2 mb-lg-0" style="
+                    padding-bottom: 0px;">
+                        <li class="nav-item" id="loader" style="
+                        padding-bottom: 0px;">
                         </li>
-                        <li class="nav-item">
-                            <a class="nav-link" aria-current="page" data-bs-toggle="modal" data-bs-target="#chanmode" type="button" onclick="populate_channel_list();">
+                        <li class="nav-item" style="
+                        padding-bottom: 0px;">
+                            <a class="nav-link" aria-current="page" data-bs-toggle="modal" data-bs-target="#chanmode"
+                                type="button" onclick="populate_channel_list();">
                                 Meetings by Channel
                             </a>
                         </li>
-                        <li class="nav-item">
+                        <li class="nav-item" style="
+                        padding-bottom: 0px;">
                             <a class="nav-link" type="button" href="#">
                                 Meetings by Group
                             </a>
                         </li>
-                        <li class="nav-item">
+                        <li class="nav-item" style="
+                        padding-bottom: 0px;">
                             <a class="nav-link" type="button" href="#">
                                 About
                             </a>
                         </li>
                     </ul>
+                    <div class="d-flex flex-column justify-content-center align-items-center">
+                        <div class="input-group">
+                            <input type="text" autocomplete="off" class="form-control"
+                                placeholder="Enter four or more characters to search"
+                                aria-label="Enter four or more characters to search" aria-describedby="findtext"
+                                style="width: 350px;" id="findtext">
+                        </div>
+                        <div class="input-group" data-bs-auto-close="true">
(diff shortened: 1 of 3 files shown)


subhangi2731:
Screenshot from 2022-01-05 23-53-33

subhangi2731:
@t0xic0der @darknao please review and let me know

t0xic0der (changes requested):
These are too many suggestions with zero context on when these meetings took place, in which channel etc. Please provide those.

image

t0xic0der (changes requested):
Also, please limit the suggestions to just the most recent five results with a button to show more results, should there be more available.

image

t0xic0der (changes requested):
These suggestions trigger only when I click on the button - which should not be the case. The search should acclimatize itself to the newly entered text so the trigger should be made whenever the text elements in the search bar change.

t0xic0der (changes requested):
Please disable the browser induced search suggestions.

Screenshot from 2022-01-06 10-50-51

t0xic0der (changes requested):
Please point the results to the meetings available "locally" with the current deployment of Mote and NOT the meetbot-raw site.

t0xic0der (changes requested):
I assume that the searchbar should be placed above the card outside of it, and NOT inside of it as that is where the calendar would be. Please make changes to the positions accordingly.

subhangi2731:
@t0xic0der Okay I will do the changes

subhangi2731:
@t0xic0der I have updated the PR and done the changes as suggested. Please review and let me know

darknao (changes requested):
Looks good. But I have a few remarks:

Your vscode settings should not be included in this PR.
The search bar indicate it's looking for meeting logs, where it actually only looks for channel name. That doesn't seems right.
Ideally, the user should be able to search either a channel name, a meeting name, or a team name.
The returning list should give some kind of context for each item type, such as an icon symbolizing if it's a channel, meeting, or team.
It could also be separated into sections, like this (taken from account.fp-o) : 
suggest

In my opinion, I think this search bar should be combined with the calendar view, as it make  it way simpler to see how frequent is a meeting, and choose a date of interest, compared to pick one in a list like it is right now.
The way I see it, the calendar view should adapt itself to show only meetings from the specific channel / meeting name / team the user is looking for, and automatically jump to the last week or month a meeting was happening with those criteria.

subhangi2731:
@t0xic0der @darknao Can you please once review the updated PR and let me know

darknao:
There is a few things that quickly come out of my mind :
- You should start requesting the backend only after 3 or 4 characters has been typed. Doing a search by string with only 1 or 2 characters will not produce any relevant results.
- If I start typing something like "ansible", the frontend will launch multiple requests, one for each letter, which is not optimal, and make the application not as responsive as it could be. Inducing a delay before making any requests can help to make sure the user is actually done typing.
- The end result may need a little bit of sorting. For instance, I would expect the latest meeting to come first on the list. Additionally, if my request return several different meetings, I rather see at least one of each, sorted by date, instead of 5 occurrences of the same meeting.

subhangi2731:
@t0xic0der @darknao please review and let me know. I did all the above changes

subhangi2731:
Screenshot from 2022-01-12 11-34-58

subhangi2731:
Screenshot from 2022-01-13 00-57-55

t0xic0der (changes requested):
Great work so far.

Some changes I'd like to suggest.

Take a close look at this image

image

Do you see that after the search bar has been added the calendar has moved on by a bit and hence in a default view, you would actually have to scroll down a bit to see the calendar in its entirety. I could ask you to shorten the height of the calendar but that would not be the right thing to do - owing that there can be days where there are more meetings and a short height will not suffice.

What would I ask you to do instead - is to move the search bar to the navbar, as the last element there beside the `About` link. The way the search operates should stay the same and only the position of it should change.

t0xic0der (changes requested):
Instruct users in the search bar placeholder (where you get to see `Search meeting logs`) to enter four or more characters to start searching for the meeting. The text can be `Enter four or more characters to search`. This is to ensure that the people are not left to imagine if the search functionality is working properly or not when they end up entering only two or three characters.

t0xic0der (changes requested):
When I click on the meeting that turn up in the suggestions, the highlight color corresponds to Bootstrap's Primary Blue. Could you please pick a lighter tint of Fedorable Blue color for which I have already shared you the palettes and resources too?

REASON - Not just because the color scheme does not match the ones we seek but it also makes the description text illegible.

Screenshot from 2022-01-13 08-50-50

t0xic0der (changes requested):
Please sort the search results which turn up in the modal in the reverse chronological order of their occurrences - the recent meetings should turn up first and the older meetings should follow after.

image

t0xic0der (changes requested):
The same goes for the meetings that turn up in the dropdown suggestions - they need to be rearranged in a reverse chronological manner of their occurrence. In the screenshot below, notice how the meetings that took place on `Jan 07 2020` appear before the ones that took place on `Jan 28 2020` - which should not be the case.

Screenshot from 2022-01-13 08-56-47

t0xic0der (dismissed):
Please ensure that the meetings open up in a new tab and not in the one that is currently housing the landing page of Mote.

subhangi2731:
@t0xic0der @darknao please review and let me know. I did the changes as said above by @t0xic0der

subhangi2731:
@t0xic0der please review and let me know. I did the changes as discussed in meet.

subhangi2731:
Screenshot from 2022-01-14 23-43-17

darknao (changes requested):
Few things:
- Please use semicolons in your javascript code, always.
- The debounce function does not work (check the scope of your variables).
- Ideally, the search function should not be triggered if the search string doesn't change. For instance, using the arrow keys, ctrl+a, or any key that doesn't alter the text in any way should not trigger a new search.
- The search function can take 1 or 2 seconds to return a result. Add some kind of visual cue for the user to know that something is happening in the background. Like a spinner somewhere.
- If the search doesn't return any results, notify the user.
  As a suggestion, instead of an empty list, you can insert the text "No meeting found" in the list. You can also use this technique to tell the user to type at least 4 characters to begin the search and keep the original placeholder text to "Search meeting logs...", which in my opinion help the user to understand the purpose of this text bar.
- In my opinion, the search bar is the core functionality of this site, and should not be hidden in the collapsible area of the navbar in mobile view. 
- Either the list width is too small, or the text is too big, but most of the meeting names are truncated and it doesn't look right.

subhangi2731:
Screenshot from 2022-01-18 23-29-49
Screenshot from 2022-01-18 23-30-04

subhangi2731:
@t0xic0der @darknao I made all the changes. Please have a look and let me know

subhangi2731:
@darknao could you please review and confirm if everything is okay ?

src/mote/templates/mainpage.html
```diff
@@ -377,8 +385,129 @@ <h6 class="modal-title head fw-bold">Recent conversations</h6>
...
+                    dropdownDiv.innerHTML =
+                        `<li><a class="dropdown-item" type="button">Loading...</a></li>`
+                    dropdownDiv.classList.add("show")
+                    prevValue = value
+                    await $.getJSON("/fragedpt/", {
+                        "rqstdata": "srchmeet",
+                        "srchtext": value
+                    }, function (data) {
+                        results = data
+                        console.log(data);
```
darknao, line 459: I assume you used that for debugging purpose. If that is the case, you can remove it if it's not needed anymore.

darknao (approved):
Looks good to me :+1:

src/mote/templates/mainpage.html
subhangi2731: sure, yes I used it for testing purpose

subhangi2731:
@darknao done

subhangi2731:
@t0xic0der Is this ready to be merged

PR #235, opened 2022-01-22
Improved calendar theme

darknao:
This is just me trying to get closer to @mairin design posted here: https://github.com/fedora-infra/mote/issues/228#issuecomment-985007664.

Dark mode included (I'm just throwing in darker colours around, I still need a proper Fedora colour scheme, if that exists).

image
image
image
image
(Those screenshots may not reflect the current state of this PR)


darknao:
That is exactly what I needed! Thank you!

t0xic0der:
Hi @darknao, (long time no see ;P). I really don't want to breath down your neck here but if you could tell me something about the current status of this PR and if you need any assistance with anything - that would be super helpful.

And @mairin, thanks for the mockups! You are as helpful as always! :D

darknao:
@t0xic0der hi :)
Last update (not pushed yet) looks like this:
image

My biggest issue right now, is figuring out the best way to get all meetings metadata (like attendee, topics, action items...) to display them as we want.
Right now, all these informations are only available in a preformatted html file, which is not ideal.
It would be nice if we could get meetbot to save these informations in a structured format like json so we can then format them as we see fit.
Same thing for datagrepper, although it's a little better on that side, as some metadata is already available there (that's where I'm getting the number of attendees from right now).
Another solution would be to scrap these data ourselves from the html minute logs, and save it back as json. But I'm not a big fan of this method as it's really not optimal.

t0xic0der:
> My biggest issue right now, is figuring out the best way to get all meetings metadata (like attendee, topics, action items...) to display them as we want.
Right now, all these informations are only available in a preformatted html file, which is not ideal.

The current version of the `meetbot` (the one that we talk to using our IRC client) unfortunately does not provide us a way to get all these information. We could scrape for these information but like you mentioned, we would rather not want to. 

> It would be nice if we could get meetbot to save these informations in a structured format like json so we can then format them as we see fit.

Would you know the right people to talk to in regards with this? This would be such a useful feature to obtain metadata information about the said meetings and that could keep us from scraping information in a less elegant manner.

> Same thing for datagrepper, although it's a little better on that side, as some metadata is already available there (that's where I'm getting the number of attendees from right now).

> Another solution would be to scrap these data ourselves from the html minute logs, and save it back as json. But I'm not a big fan of this method as it's really not optimal.

Let me come up with something. I will be a while because when using `regex` to solve ONE problem, we actually end up creating TWO problems but if we are able to `BeautifulSoup4` our way through it - that should be cool. What do you think?

darknao:
> Would you know the right people to talk to in regards with this? This would be such a useful feature to obtain metadata information about the said meetings and that could keep us from scraping information in a less elegant manner.

I think meetbot is still maintained by @ryanlerch iirc.

darknao:
I'm getting there slowly...
image 
image

I took the bs4 route, and surprisingly, it's not that bad performance wise.

copperii:
Last picture, it should say x people, not x peoples.

t0xic0der (approved):
LGTM.

Thanks for all the great work!

PR #236, opened 2022-01-28
Render template of teams and meetings data from yaml file

subhangi2731:
I created a separate YAML file where I have listed all the teams and their associated meeting names. Then I created a new function which has read the team yaml file and then returned a new html page with every data in it. Thus to send html content, I used the function "render_template" where the first argument of this function is the teams.html template file which I created in the template directory. Used @main.get annotation with teams URL to get all the teams data.


subhangi2731:
Screenshot from 2022-01-28 19-44-24

subhangi2731:
@darknao @t0xic0der please have a look to it and let me know if any corrections are needed

subhangi2731:
Screenshot from 2022-02-03 12-23-52
Screenshot from 2022-02-03 12-23-44
Screenshot from 2022-02-03 12-23-25

subhangi2731:
@t0xic0der @darknao please check and let me know how it looks

src/mote/main.py
```diff
@@ -218,5 +221,17 @@ def page_not_found(error):
...
 
+@main.get("/teams")
+def teamspage():
+    path = os.path.abspath("src/mote/teams.yml")
+    data = dict()
+    with open(path, "r") as stream:
+        try:
+            data = yaml.safe_load(stream)
+        except yaml.YAMLError as exc:
+            print(exc)
```
darknao, line 232: Use the python logging facility instead of print. Something like 
```python
logging.error("Your message")
```

src/mote/templates/teams.html
```diff
@@ -0,0 +1,380 @@
...
+    <link href="{{ url_for('static', filename='css3/fontawesome-all.min.css') }}" rel="stylesheet" />
+    <link href="{{ url_for('static', filename='css3/fullcalendar.css') }}" rel="stylesheet" />
+    <script src="{{ url_for('static', filename='jscn/bootstrap.bundle.min.js') }}"></script>
+    <script src="{{ url_for('static', filename='jscn/jquery.min.js') }}"></script>
+    <link rel="icon" href="{{ url_for('static', filename='imgs/logoteal.png') }}" sizes="16x16">
+    <script src="{{ url_for('static', filename='jscn/fragment.js') }}"></script>
+    <script src="{{ url_for('static', filename='jscn/easy.qrcode.min.js') }}"></script>
+    <script src="{{ url_for('static', filename='jscn/socket.io.js') }}"></script>
+    <script src="{{ url_for('static', filename='jscn/fullcalendar.js') }}"></script>
+    <link href="{{ url_for('static', filename='css3/teams.css') }}" rel="stylesheet" />
```
darknao, line 39: We don't really need a separate css file for this page. You should merge this one with the main one `fragment.css`.

src/mote/templates/teams.html
```diff
@@ -0,0 +1,380 @@
...
+                </div>
+            </div>
+        </div>
+    </nav>
+    <div class="container-lg">
+        <div class="card mt-1" id="bigcard">
+            <div class="card-body pt-0">
+                <div class="mt-4">
+                    <br><br>
+                    <article>
```
darknao, line 97: I'm not sure why you choose to use the `article` tag here. A simple `div` should be more than enough.

src/mote/templates/teams.html
```diff
@@ -0,0 +1,380 @@
...
+                    <a style="color:#676767; text-decoration:none;"
+                        href="https://github.com/t0xic0der/fragment/graphs/contributors" target="_blank">
+                        others
+                    </a>
+                </span>
+            </div>
+        </div>
+    </div>
+</body>
+<script>
```
darknao, line 162: Most of this is related to the calendar plugin, which is not on this page.
It should probably be removed.

src/mote/main.py
```diff
@@ -21,11 +21,14 @@
...
 
 import logging
+import os
 import re
 from datetime import datetime
 from threading import Event, Lock
 
 import click
+import flask
+import yaml
```
darknao, line 31: This must be added as a dependency in `requirements.txt`

src/mote/templates/teams.html
```diff
@@ -0,0 +1,380 @@
...
+        <div class="card mt-1" id="bigcard">
+            <div class="card-body pt-0">
+                <div class="mt-4">
+                    <br><br>
+                    <article>
+                        <i class="fa fa-users" aria-hidden="true"></i>
+                        Meetings by Group
+                        <input class="form-check-input" type="checkbox">
+                        Hide inactive
+                    </article>
```
darknao, line 102: While it may look good with your screen width, the alignment breaks with either a narrow or wider window.
Check this https://getbootstrap.com/docs/5.0/utilities/flex to see how to achieve proper alignment in a responsive way.
There is several examples in this page that should help you do that.

darknao (changes requested):
Sorry it took me so long to review this...

For reference purpose: link to Marin design https://user-images.githubusercontent.com/799683/144504408-0031c0d5-684b-4e41-ba25-93942719489c.png

See if you can take a little step closer to this design when it comes to font choice, size, and color of the various elements in the team list.

subhangi2731:
@t0xic0der @darknao done all the changes. Please have a look and let me know

src/mote/templates/teams.html
```diff
@@ -0,0 +1,314 @@
...
+                                <tr>
+                                    <td class="item">{{i}}</td>
+                                    <td class="item">
+                                        <item>
+                                            Latest Minutes:
+                                        </item>
+                                        <div class="btn-group" role="group"
+                                            aria-label="Basic checkbox toggle button group">
+                                            <input type="checkbox" class="btn-check" id="btncheck1" autocomplete="off">
+                                            <label class="btn btn-outline-primary" for="btncheck1"><i
```
darknao, line 121: All buttons will share the same id, which is not right.
Either make this ID unique, or you could simply remove it since you're not using it anyway.

src/mote/templates/teams.html
```diff
@@ -0,0 +1,314 @@
...
+                        <div><input class="form-check-input" type="checkbox"> Hide inactive</div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+                    <br><br>
+                    <div class="card">
+                        <div class="card-header">
+                            23 Groups
+                        </div>
```
darknao, line 108: That would be super nice if the number of groups was not static.
Try to find a way to get this number from the data you already have.

src/mote/templates/teams.html
subhangi2731: sure

src/mote/templates/teams.html
subhangi2731: sure

src/mote/templates/teams.html
```diff
@@ -0,0 +1,314 @@
...
+                        <div><i class="fa fa-users" aria-hidden="true"></i> Meetings by Group</div>
+                        <div><input class="form-check-input" type="checkbox"> Hide inactive</div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+                    <br><br>
+                    <div class="card">
+                        <div class="card-header">
+                            {{length}} Groups
```
darknao, line 107: I guess that work. But you don't really need to define another variable, you can use jinja2 filter instead:
```suggestion
                            {{ teamdata|length }} Groups
```

subhangi2731: done

mote/main.py
```diff
@@ -21,10 +21,13 @@
 """
 
 import logging
+import os
 import re
 from datetime import datetime
 
 import click
+import flask
+import yaml
```
darknao, line 30: Duplicate import, please remove it.

mote/main.py
```diff
@@ -215,5 +218,17 @@ def page_not_found(error):
...
 
+@main.get("/teams")
+def teamspage():
+    path = os.path.abspath(main.config["TEAMS_FILE_LOCATION"])
+    data = dict()
+    with open(path, "r") as stream:
+        try:
+            data = yaml.safe_load(stream)
+        except yaml.YAMLError as exc:
+            logging.error("exc")
```
darknao, line 229: If your goal was to log the exception, the correct syntax should be:
```suggestion
            logging.error(exc)
```

mote/static/css3/fragment.css
```diff
@@ -138,4 +138,55 @@ hr {
...
+
+#bigcard {
+    border: solid transparent;
+  }
+  .card li {
+    color: rgba(60, 110, 180, 1);
+    font-family: 'Montserrat';
+    font-weight: bold;
+
+  }
```
darknao, line 151: Please use the same indent as existing content in this file.

darknao (changes requested):
Now that you have rebased your work, please remove `requirements.txt` and add all missing required libraries to `pyproject.toml`

subhangi2731:
@darknao @t0xic0der I have done all the changes. Is this ready to be merged?

darknao:
Not yet.
All checks are failing and needs to be fixed.
`poetry.lock` needs to be updated to include PyYAML. You can use `poetry lock --no-update` for that.
Note that you should have used the `poetry add PyYAML` command in the first place instead of editing the `pyproject.toml` file by hand. This would have updated both files for you.

And the `yaml` import is missing from `main.py`.

subhangi2731:
@darknao @t0xic0der resolved and updated all the dependencies in poetry.lock file. The checks are passing now. Now is this ready to be merged?

darknao (approved):
Looks good :+1:

PR #313, opened 2022-06-26
use standard fedora favicon

darknao:


t0xic0der (approved):
LGTM.

(P.S. Just a nitpick - feel free to disregard but I would suggest moving out the changes made to the Dockerfile from this PR as they do not relate much to changing the standard Fedora favicon.)

t0xic0der:
Thanks. Merging!

Issue #321, opened 2022-06-28
Minutes in text format are 404


darknao:
zodbot returns a set of 3 links at the end of each meeting: Minutes in html, minutes in text, and the log:
```
Meeting ended Mon Jun 27 18:26:12 2022 UTC. Information about Zodbot's MeetBot at https://fedoraproject.org/wiki/Zodbot#Meeting_Functions. 
Minutes: https://meetbot.fedoraproject.org/fedora-meeting-3/2022-06-27/fedora_infrastructure_ops_daily_standup_meeting.2022-06-27-18.00.html
Minutes (text): https://meetbot.fedoraproject.org/fedora-meeting-3/2022-06-27/fedora_infrastructure_ops_daily_standup_meeting.2022-06-27-18.00.txt
Log: https://meetbot.fedoraproject.org/fedora-meeting-3/2022-06-27/fedora_infrastructure_ops_daily_standup_meeting.2022-06-27-18.00.log.html
```
The text one is not working anymore with the new version of meetbot.

PR #323, opened 2022-06-28
A couple of fixes

darknao:
For testing purposes, we are currently downloading new meeting files from meetbot-raw when notified by fedora-messaging.
Now that we are in production, we don't need to do that anymore as we already have direct access to all meeting files locally.
Additionally, the filesystem is read-only, so the update logic is failing and prevent the calendar view from showing any new meeting.

The second commit is fixing #321 by redirecting the user to meetbot-raw when requesting meeting logs in (raw) text format.

### DIFF
--- a/mote/main.py
+++ b/mote/main.py
@@ -24,7 +24,7 @@ import re
 from datetime import datetime
 
 import click
-from flask import abort, jsonify, render_template, request, url_for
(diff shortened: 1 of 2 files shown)


t0xic0der (approved):
LGTM.

PR #335, opened 2022-07-05
improve darktheme for log view

### LINKED ISSUE #322: Meeting logs in dark mode are hard to read
The color scheme used in the darkmode theme should be changed to improve readability:
image

darknao:
before / after:
image
image

light mode for reference:
image

Thoughts?

Fix #322

### DIFF
--- a/mote/static/css3/fragment.css
+++ b/mote/static/css3/fragment.css
@@ -51,6 +51,10 @@ pre {
     white-space: pre-wrap;
 }
 
+[data-bs-color-scheme="dark"] pre  {
+    color: #ffffff;
+}
+
 .dropdown-item:focus {
     background-color: #108BD6;
 }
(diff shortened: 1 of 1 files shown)


t0xic0der (approved):
LGTM.

I will let this PR standby for around 12 more hours as I pinged more folks regarding their opinion on this. If we do not get any more change requests on this by then, we will merge this.

lilyx13:
It looks much better with this color scheme. I would also suggest to also change the light theme time font color to be grey instead of green for more consistency as well.

grahamwhiteuk:
That looks considerably better to me!  So much easier to read, good work.

darknao:
new set:
image
image
switch timestamps color to grey in light mode, and more white-ish  in dark mode.

Issue #342, opened 2022-07-11
RFE: text links in the new UI


mattdm:
The new UI looks great! Thanks for doing this.

I often want the text-formatted version of the minutes — it's handy for re-posting into email or forum posts.

I _think_ this just applies to the minutes, because the "full log" view contains a basically-raw log anyway that one can copy. But maybe a link from _that_ page to the corresponding text would be nice too.

I happen to know that I can click on "Full Logs" or "Permalink" and then change the URL to end with `.txt` instead of `.html`. But that's obscure, and kind of clunky and annoying even if you do know it.

I don't know if the UI answer is "MORE BUTTONS!" -- maybe the "permalink" button could become one of those buttons which works if you just click it but also has a drop-down with more options (in this case, "plain text").

darknao:
Something like this?
![image](https://user-images.githubusercontent.com/693402/178341115-dfd52e24-aa35-4f4a-a103-7b890b14fd26.png)

PR #344, opened 2022-07-11
Add links for plain text logs/minutes (#342)

darknao:
Add dropdowns to [Full Logs] and [Permalink] buttons to select between html and text version.


t0xic0der (approved):
LGTM.

PR #345, opened 2022-07-11
avoid running duplicate jobs on PR

darknao:
In its current configuration, CI jobs are run twice on PRs from feature branch (or dependabot ones): 
- first one for the `push` event on said branch
- and second one for the `pull_request` event toward the `main` branch
This PR should prevent that.


t0xic0der (approved):
LGTM.

Issue #338, opened 2022-07-07
RFE: search for meetings matching a name


keszybz:
[from https://discussion.fedoraproject.org/t/meetbot-logs-2-0-is-out-now/40427/3]

RFE: list all meetings where the title matches some pattern, ordered by date.

For "fesco" this almost works, but the results are not ordered by date, so it's hard to look for a specific entry. Also, the list is limited to a few entries, so again, if I wanted to get a list of all fesco meetings over the last year, it'd be very hard with this interface. 

Background: occasionally I need to access logs for a specific meeting, for example when I want to link the logs to a meeting from before a few weeks into some ticket. For example, let’s say that I want to look for FPC meeting logs over the last year. One option is to go through the calendar and look for a meeting and start from there. But let’s say that I don’t know when the last meeting took place and it only happens every month or so, so that’d be inefficient. So I want to try search. But “FPC” is just three characters, so I can’t search for that. And searching for “Packaging” or “Committee” yields seemingly random entries in random order.

keszybz:
![Screenshot from 2022-07-14 19-08-37](https://user-images.githubusercontent.com/349618/179042468-5645dc59-03d9-4ab0-b229-a679115ff77f.png)

With "more":
![Screenshot from 2022-07-14 19-10-26](https://user-images.githubusercontent.com/349618/179042683-4aae9149-eea8-4e5a-b01f-e58eaff5bdf4.png)

darknao:
ah, interesting.
Can you share some information about your environment, like operating system, browser, and the locale setting (language used by your browser)?

keszybz:
I get the same search order with google-chrome-stable-98.0.4758.102-1.x86_64.

darknao:
Thanks, that is very helpful.
I may have found something. Can you try https://mote-mote.apps.ocp.stg.fedoraproject.org/ and see if you've the same issue.
Please keep in mind that this environment has a much smaller dataset, most of the meetings logs won't be there, but it should be enough for testing purposes.

keszybz:
With  https://mote-mote.apps.ocp.stg.fedoraproject.org/, I get the expected order!

darknao:
Wonderful! Thanks for confirming.
The fix should land in production in the next days.

Thanks for your help.

PR #351, opened 2022-07-14
lower search restriction to 3 characters

darknao:
Related to #338 
Some meetings name like `FPC` are too short and cannot be looked up with the current setting.


t0xic0der (approved):
LGTM.

PR #353, opened 2022-07-14
use iso8601 format for date in backend for reliability

### LINKED ISSUE #338: RFE: search for meetings matching a name

darknao:
Uses ISO8601 date format when needed to make date conversion reliable and locale independent.
Also uses timestamp to sort meetings in search results.
Should fix #338.

### DIFF
--- a/mote/modules/find.py
+++ b/mote/modules/find.py
@@ -70,8 +70,6 @@ def find_meetings_by_substring(search_string: str):
                     location_list[1],
                     location_list[2],
                 )
-                formatted_timestamp = datetime.strptime(meeting_date, "%Y-%m-%d")
-                datestring = "{:%b %d, %Y}".format(formatted_timestamp)
                 if ".log.html" in meeting_filename:
                     meeting_log_filename = meeting_filename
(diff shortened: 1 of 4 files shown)


t0xic0der (approved):
LGTM.

PR #380, opened 2022-08-09
Cleaned up and fixed some fonts, you see ;P

t0xic0der:
Before:

image

After:

image


Zlopez (approved):

darknao (approved):
Looks good to me.
Just for information, this is how it looks on my computer (Fedora 36 / Firefox 102):
Before (in prod):
image

After (this change):
image

On my side, the changes are not that big, but maybe that's just me.
If in the end, we both end up with pretty much the same renderering, I consider that an improvement :)

t0xic0der:
@darknao, I tested the fonts available on the production deployment on a Windows 10 PC (Please don't fire me). That system did not have Open Sans fonts, to begin with, which is why it settled with the next best font face available (Ref. https://github.com/fedora-infra/mote/blob/main/mote/static/css3/fragment.css#L33). I am going to assume that you are testing this on a Fedora Linux PC and that is why it is settling for the Open Sans fonts because your system can provide those fonts that were (previously) unavailable (Ref. https://github.com/fedora-infra/mote/tree/main/mote/static/font) from the server.

Issue #214, opened 2021-11-03
Make more uses of Jinja templates


darknao:
We only have 3 html templates for now, but we already have parts of the same code on all of them.

Use the power of templating to refactor those pages and eliminate code duplicates like footer & header.
See https://jinja.palletsprojects.com/en/3.0.x/templates/

subhangi2731:
@darknao can I work on this?

t0xic0der:
@subhangi2731, there you go.

subhangi2731:
@t0xic0der @darknao As I've observed, this needs topbar & footers across all the pages to be consistent with each other, which isn't the case currently. So, this issue can be worked only on merging these two short MRs: #207 #210

PR #381, opened 2022-08-09
Add project repo information in the footer

### LINKED ISSUE #363: As a prospective contributor, I want to quickly find the link to mote sources on meetbot.fp.o
I was looking around meetbot.fp.o for a link to the sources (here). I couldn't find one. Could one please be included in the footer, and maybe on the "about" page too?

t0xic0der:
Signed-off-by: Akashdeep Dhar <akashdeep.dhar@gmail.com>

Fixes #363

@sanjayankur31, could you please take a look at this?

Before:

image

After:

image

### DIFF
--- a/mote/templates/aboutpage.html
+++ b/mote/templates/aboutpage.html
@@ -487,6 +487,10 @@
                            target="_blank">
                             Fedora Websites and Apps Team
                         </a>
+                        <a href="https://github.com/fedora-infra/mote"
+                           target="_blank">
+                            (Sources)
+                        </a>
                     </span>
                 </div>
             </div>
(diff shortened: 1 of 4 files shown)


sanjayankur31:
lgtm :+1:, thanks! :)

Zlopez (approved):
Looks good, but it would be better to have some master template that will contain header and footer, so you don't need to change it in multiple places.

darknao (approved):
Looks good.
Agree with @Zlopez: a master template for the footer should be best but can be done on a separate PR (we have a ticket for that #214).
I would personally put the link right on the "Meetbot Logs" text instead, but that's just my opinion, it's fine that way too.

PR #386, opened 2022-08-11
Rename templates into J2 extension and correct references

t0xic0der:
Rename templates into J2 extension and correct references

No, this does not address #214 but makes it easy to do so.


darknao (approved):
LGTM

Zlopez (approved):

PR #387, opened 2022-08-11
Fix #385: event summary modal takes full width in full screen

### LINKED ISSUE #385: Occupy the entire width of the modal for the content
This is how it is right now 

image

A big chunk of the modal real-estate is left unutilized and that should not be the case.

darknao:


mote/templates/event_summary.html
```diff
@@ -30,7 +30,7 @@
   </div>
 </div>
 
-<div class="evt-smry-body container p-3 pb-0">
+<div class="evt-smry-body container-fluid p-3 pb-0">
```
t0xic0der, line 33: Hi @darknao, could you please resolve the merge conflicts which were introduced as a result of #386 getting merged?

t0xic0der (changes requested):

mote/templates/event_summary.html
darknao: done

t0xic0der (approved):
LGTM.

PR #446, opened 2022-10-22
add OGP meta tags

darknao:
Make this:
image

Into this:
image

### DIFF
--- a/mote/templates/base.html.j2
+++ b/mote/templates/base.html.j2
@@ -23,7 +23,9 @@
     <head>
         <meta charset="utf-8"/>
         <meta name="viewport" content="width=device-width, initial-scale=1"/>
-	<meta name="color-scheme" content="light dark">
+        <meta name="color-scheme" content="light dark">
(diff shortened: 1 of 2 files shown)


t0xic0der:
Hi @darknao, I want to make one small suggestion. 

As the project is (unironically) called Meetbot Logs, it would be rather helpful to call the Logs as "Meetbot Logs - Logs" and Minutes as "Meetbot Logs - Minutes" to ensure the consistency between the actual project name and the functional name of the page.

darknao:
@t0xic0der: sounds fair. And to avoid repetition, I propose this middle ground: Just use "Meetbot Logs" as the title. The type of logs (logs or minutes) is already mentioned in the description line just below.

darknao:
After some thought, `og:title` is supposed to be for the content title as opposed to the website name.
For the site name, there is actually an `og:site_name` tag available.
Here is how it looks:
image

t0xic0der (approved):
LGTM

PR #447, opened 2022-10-22
add ability to browse adjacent meetings

### LINKED ISSUE #364: As a Fedora community member, I'd like to have a permalink to the last meeting that a team held, or that was held in a channel
Most teams do look at tasks from the last meeting, and lots of us would like to quickly find the latest meeting in general to go through logs. So, it'll be very useful for us to have a permalink for the most recent meeting. 

For example, it could be: https://meetbot.fedoraproject.org/fedora-neuro/latest, and would point to the latest meeting that was held in the fedora-neuro channel.

If this can also be done for teams or meeting names, that would be even nicer (since some meeting rooms are used by multiple teams):

https://meetbot.fedoraproject.org/neurofedora/latest

darknao:
This was sitting on staging for quite some time now.
I think it's time to push it on prod.

Close #364

### DIFF
--- a/mote/main.py
+++ b/mote/main.py
@@ -22,6 +22,7 @@
 
 import re
 import urllib.parse
+from datetime import datetime
 
 import click
 from flask import abort, jsonify, redirect, render_template, request, url_for
@@ -29,6 +30,7 @@ from flask import abort, jsonify, redirect, render_template, request, url_for
 from mote import app as main
 from mote import logging, socketio
 from mote.__init__ import __version__
+from mote.modules import sanitize_name
 from mote.modules.call import (
     fetch_channel_dict,
     fetch_datetxt_dict,
@@ -36,7 +38,7 @@ from mote.modules.call import (
     fetch_meeting_dict,
     fetch_meeting_summary,
 )
-from mote.modules.find import find_meetings_by_substring
+from mote.modules.find import find_meetings_by_substring, get_meeting_adj
 from mote.modules.late import fetch_meeting_by_period
 
 thread = None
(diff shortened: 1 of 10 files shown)


t0xic0der:
Hi @darknao,

Are the instructions provided in the README good enough to test this PR out or do I need to do something else apart from that?

Owing to the scale of this PR, I could use another pair of eyes to review this. @onuralpszr, you're up :D

darknao:
> Are the instructions provided in the README good enough to test this PR?

Yes, they are.

t0xic0der (approved):
Works as intended.

When I click on a calendar entry, I kinda have to wait for some time for the modal to appear. As the loading is not indicated, the users are likely to try clicking on other calendar entries to check if they work and that leads to multiple modal windows opening up and getting dismissed by the modal windows that follow them. But after the initial wait for some ~10 seconds, it works fast enough for the opened set of meetings. 

image

I previously thought of it as a minor gripe but I think that if the caching takes the aforementioned amount of time on a local setup, chances are that it might take more when accessed remotely. I might be mistaken here too as we are using Redis for caching purposes in the production and that can help alleviate the issues of delays faced due to caching adjacent meetings of the same set.

image

darknao:
> But after the initial wait for some ~10 seconds, it works fast enough for the opened set of meetings.

I don't think this is related to this PR, but ~10 seconds seems very concerning and I never experienced such a long delay to open the modal, especially on a local deployment with a limited data set.
Would you mind opening a new ticket regarding this, I'll take a closer look at what we can do to make the user experience a bit better.

t0xic0der:
> I don't think this is related to this PR, but ~10 seconds seems very concerning

I would hope so too.

I am merging this one for now and wait for more such similar reports while opening up a ticket on the same from my end.

